PR #4, opened 2020-03-16
Add simple .travis.yml for CI testing

eriknw:
Jim, you'll need log in to https://travis-ci.org/ and enable CI for this repo (under Settings->Repositories I think).

This isn't quite working yet.  `python setup.py build_ext` or `pip install -e .` fail.  See: https://travis-ci.org/github/eriknw/grblas/jobs/663215948

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,10 @@ __pycache__/
 # C extensions
 *.so
 
+# Vi
+*.swp
+*.swo
+
 # Distribution / packaging
 .Python
 build/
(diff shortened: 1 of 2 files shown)


jim22k:
Not sure why it can't find GraphBLAS.h. That's available in graphblas, which we're pulling from conda-forge.

When I make the conda recipe, I did have to put:
```
$PYTHON setup.py build_ext -I $PREFIX/include -L $PREFIX/lib
```
in build.sh. Maybe something similar is needed for travis to be happy.

eriknw:
Good idea!  Looks like using `$CONDA_PREFIX` worked.  This doesn't feel like the best way, but a working CI is better than a non-working CI.  We may need to do more later if e.g. we want to support installing via pip.

Issue #5, opened 2020-03-16
Change Operator access


jim22k:
## Current Operator access
```python
from grblas import UnaryOp, BinaryOp, Monoid, Semiring

UnaryOp.MINV
BinaryOp.PLUS
Monoid.TIMES
Semiring.MIN_PLUS
```

## Proposed Operator access
```python
from grblas import unary, binary, monoid, semiring

unary.min
binary.plus
monoid.times
semiring.min_plus
```

## Details
There will still be `grblas.ops` which contains `UnaryOp`, `BinaryOp`, `Monoid`, `Semiring`. These classes will be the underlying type of the operators. Registering new user operators will be done from these classes, but the name chosen will become a new variable in the associated module.

The associated modules (`unary`, `binary`, `monoid`, `semiring`) will only contain instances of operators (builtin or user-defined).

I want to attempt to remove the current mechanism of regex matching. Making an explicit table of operators and their mapping to GraphBLAS will be clearer, even if it's more verbose than the current approach.

PR #6, opened 2020-03-16
Move operator instances to own module

jim22k:
See Issue #5 for the rationale behind these changes.

`BinaryOp.PLUS` becomes `binary.plus`

`binary.plus` is still an instance of `BinaryOp`, but we make instances available in their own module rather than bolting them on to the class.


eriknw:
LGTM

PR #7, opened 2020-04-16
Structural mask

jim22k:
SuiteSparse::GraphBLAS added the structural mask from spec v1.3. This updates `grblas` to use the structural mask via `mask.S`. The complemented structure is given by `~mask.S`.

Other improvements:

- Add `PAIR`, `ANY`, `ONE`, `ABS` operators
- New `isequal` method for comparing two objects, with parameters to control tolerance of floating point comparison and whether to require strict dtype matching


grblas/matrix.py
```diff
@@ -22,20 +22,43 @@ def __repr__(self):
         return f'<Matrix {self.nvals}/({self.nrows}x{self.ncols}):{self.dtype.name}>'
 
     def __eq__(self, other):
-        # Borrowed this recipe from LAGraph
+        return self.isequal(other)
+
+    def isequal(self, other, rel_tol=1e-7, abs_tol=0.0, strict_dtype=False):
```
eriknw, line 27: I'm not a fan of defaulting `isequal` to do `isclose`.  Equals means equals to me and should correspond to a single GraphBLASS call.  Why not also have `isclose`?  In numpy, `isclose` is element-wise and `allclose` returns boolean True or False, so I suppose there could be some confusion in behavior of a like-named function.

I've probably asked this before, but what do you think about adding element-wise operations such as `__lt__`?  It begs the question whether `ewise_add` or `ewise_mult` should be used, but `ewise_mult` is the only one that makes sense.  Defining `__eq__` as above that returns boolean would make adding these in the future more difficult.  Adding element-wise operations as conveniences also impacts our name of `isclose` or `allclose` if we were to go that route.

I'm curious whether and when the non-blocking version of SS:GraphBLAS will be able to fuse the operations used for the tolerance calculation and then short-circuit when a False is encountered during the `LAND` reduction.  Actually, never mind, we need to calculate the number of values from the comparison, so we can't delay.  Relatedly, I'm curious how a numba function to perform the tolerance calculation would perform.

I'm also split on using `rel_tol=` and `abs_tol=` vs `rtol=` and `atol=`.  The former is used by `math.isclose` and the latter by `numpy.isclose`.  Where did you come up with the default tolerance values?  I slightly favor `rtol` and `atol` (probably because I'm more familiar with numpy).

I'm not crazy about `strict_dtype=` either, but I can live with it.  `A.dtype == B.dtype and A.isequal(B)` isn't that bad, but neither is `A.isequal(B, strict_dtype=True)`.  Also, `strict_dtype` should be keyword-only IMHO.

grblas/backends/suitesparse/clean_header.py
```diff
@@ -4,7 +4,7 @@
 
 
 def rename_defined_constants(text):
-    #define GB_PUBLIC extern
+    # #define GB_PUBLIC extern
```
eriknw, line 7: Hahaha!

grblas/base.py
```diff
@@ -82,7 +86,7 @@ def __call__(self, *optional_mask_and_accum, mask=None, accum=None, replace=Fals
         for key in optional_mask_and_accum:
             if isinstance(key, GbContainer):
                 mask_arg = key
-            elif type(key) is ComplementedMask:
+            elif type(key) in (ComplementedMask, StructuralMask, ComplementedStructuralMask):
```
eriknw, line 89: Small nit: I would use a set literal here.  I think Python is pretty good about making it fast.  A year or two ago I would have used a tuple as well.

grblas/matrix.py
```diff
@@ -22,20 +22,43 @@ def __repr__(self):
...
+        else:
+            common_dtype = dtypes.unify(self.dtype, other.dtype)
+
         matches = Matrix.new_from_type(bool, self.nrows, self.ncols)
-        matches << self.ewise_mult(other, binary.eq)
+        # Special handling for floating point comparisons
+        if common_dtype in (dtypes.FP32, dtypes.FP64):
+            tmp1 = self.apply(unary.abs).new(dtype=common_dtype)
+            tmp2 = other.apply(unary.abs).new(dtype=common_dtype)
+            tmp1 << tmp1.ewise_mult(tmp2, monoid.max)
```
eriknw, line 53: We can compare `self.nvals` to `tmp1.nvals` here and return False if unequal.

grblas/matrix.py
jim22k: Per the discussion, let's go with:
`A.isequal(B, *, check_dtype=False)`
`A.isclose(B, rtol=1e-7, atol=0.0, *, check_dtype=False)`
Remove `==`

eriknw: I've been going back and forth on `numpy`-like vs `math`-like.  Right now I prefer:

`A.isclose(B, *, rel_tol=1e-9, abs_tol=0.0, check_dtype=False)`

which matches the signature of `math.isclose`.

grblas/matrix.py
```diff
@@ -202,109 +251,130 @@ def new_from_values(cls, rows, columns, values, *, nrows=None, ncols=None, dup_o
...
+                      binary.gt where left=Missing and right=0 yields False
+            The behavior is caused by grabbing the non-empty value and using it directly without performing
+            any operation. In the case of `gt`, the non-empty value is cast to a boolean.
+            For these reasons, users are required to be explicit when choosing this surprising behavior.
         """
         if not isinstance(other, Matrix):
             raise TypeError(f'Expected Matrix, found {type(other)}')
-        if op is NULL:
+        if op is None:
             op = binary.plus
```
eriknw, line 273: Should this be `monoid.plus`?  Will `binary.plus` work if `require_monoid=True`?

jim22k: Yes it should. I fixed it for Vector, but forgot for Matrix. Good eye!

grblas/base.py
```diff
@@ -160,7 +161,9 @@ def _update(self, delayed, mask=NULL, accum=NULL, replace=False):
         elif isinstance(accum, ops.BinaryOp):
             accum = accum[self.dtype]
         elif type(accum) is ffi.CData and ops.find_opclass(accum) != ops.UNKNOWN_OPCLASS:
-            pass
+            opclass = ops.find_opclass(accum)
+            if opclass != ops.BinaryOp.__name__:
+                raise TypeError(f'accum must be a BinaryOp, not {opclass}')
```
eriknw, line 166: Why can't `accum` be a monoid?

grblas/mask.py
```diff
@@ -0,0 +1,63 @@
...
+        self.mask = mask
+
+    def __repr__(self):
+        return f'{self.__class__.__name__} of {self.mask}'
+
+
+class ComplementedMask(Mask):
+    complement = True
+    structure = False
+    value = False
```
eriknw, line 16: No need to set the `False` values here, or no need to set all the attributes to `False` in the `Mask` base class.

grblas/base.py
jim22k: That's how the spec is written. Accumulator is always a BinaryOp -- Monoids are not allowed.

I supposed we could get fancy and extract the binaryop out of the monoid for UDFs, but for built-in monoids, that would be harder. I think it's fine to require binaryop. The error message tells the user how to fix the issue.

grblas/mask.py
jim22k: I know, but I wanted to make it easier to see what each type of mask will do.

grblas/vector.py
```diff
@@ -20,21 +20,70 @@ def __del__(self):
...
+        if check_dtype:
+            # dtypes are equivalent, so not need to unify
+            common_dtype = self.dtype
+        else:
+            common_dtype = dtypes.unify(self.dtype, other.dtype)
+
+        matches = Vector.new_from_type(bool, self.size)
+        tmp1 = self.apply(unary.abs).new(dtype=common_dtype)
+        tmp2 = other.apply(unary.abs).new(dtype=common_dtype)
+        tmp1 << tmp1.ewise_mult(tmp2, monoid.max)
```
eriknw, line 77: Hmm, I bet these two lines (76-77) could be replaced with
```python
        tmp1(mask=other.S, accum=binary.max, replace=True) << other.apply(unary.abs)
```
Is this better?  I don't have much intuition yet on performance or behavior.  We would still need the second temporary to calculate `abs(a - b)`.

jim22k: Yeah, this would be more efficient. Let me update it.

eriknw: We could then define `tmp2` below via
```python
        tmp2 = self.ewise_mult(other, binary.minus).new(dtype=common_dtype, mask=tmp1.S)
```
Once again, I don't have much intuition here.  Providing the mask here is unnecessary, but will it help or hinder performance?

jim22k: Oh wait, this won't work. If an element only exists in `other`, it will exist in `tmp1`. To make this work, we would need a mask of `self.S & other.S`, which probably defeats the effort of avoiding tmp2.

grblas/base.py
eriknw: Oh okay.  Sticking to the spec is good.  I know the spec is sometimes flexible in what is allowed.

grblas/vector.py
eriknw: Oh, right.  Is there a test that would catch this?

jim22k: Just added tests for this

grblas/base.py
```diff
@@ -242,8 +258,9 @@ def new(self, *, dtype=None, mask=None):
         if mask is None:
             output.update(self)
         else:
-            if not isinstance(mask, output.__class__):
-                raise TypeError(f'Mask must be type {output.__class__}')
+            if isinstance(mask, Mask):
+                if not isinstance(mask.mask, output.__class__):
+                    raise TypeError(f'Mask object must be type {output.__class__}')
```
eriknw, line 263: Heh, why not use `and`?

grblas/mask.py
eriknw: At a glance, I would find it easier to read if only the `True` values were specified.  Your call.

eriknw:
LGTM!

PR #9, opened 2020-04-17
Let `new_from_values` work with empty values.

eriknw:
Although silly--since you should probably use `new_from_type`--this is nevertheless a valid spelling that can come up when experimenting.

I bumped into this when I shied away from `new_from_type`, because I thought to myself "I don't have a type to get a new thing from".  What do you think of `new_from_dtype` or simply `new`?


eriknw:
Oh, note that this merges to #7 since I use `isequal`

jim22k:
I like the direction of this PR.

Now that we allow `.dup()`, the use of `Vector.new_xxx` isn't the only way to create a new object.

The spec was always a little strange because you had to first call `Vector_new` and then `Vector_build`. I don't want to require two calls and I like the ability to infer the size and type. But maybe we should get closer to the spec by modifying our methods.

#### Old methods
- Vector.new_from_type(dtype, size)
- Vector.new_from_values(indexes, values, *, size=None, dtype=None)
- Vector.new_from_existing(other)
- v.rebuild_from_values(indexes, values)
- v.dup()  _(added recently)_

#### New methods
- Vector.**new**(dtype, size)
- Vector.new_from_values(indexes, values, *, size=None, dtype=None)
- Vector.new_from_existing(other)  _(should we deprecate this?)_
- v.**build**(indexes, values)
- v.dup()

This shortens our method names and allows for someone to write `Vector.new(dtype, size).build(indexes, vals)` if they really want. No harm in that.

Is there a potentially shorter name for `new_from_values`? My first attempt is to use `Vector.build`, but that might be confusing to users that there is both a class method and an instance method called build which do something nearly the same, but not quite.

The original idea behind `rebuild` was that it can auto-clear and resize, but I think it will be used rarely enough that we can ask users to be explicit by `existing.clear().build(new_idx, new_vals)`.

PR #10, opened 2020-04-18
Add `register_anonymous` to BinaryOp and use this for `isclose`.

eriknw:
Oops, I still need to do this for `UnaryOp`.

We cache the compilation of `isclose` based on `(abs_tol, rel_tol)`.  Overall, I don't know the performance difference of this `isclose`, but just as importantly this should be more memory efficient.

This is still a little rough, but I put this up for feedback.


grblas/ops.py
```diff
@@ -191,10 +191,9 @@ class BinaryOp(OpBase):
     all_known_instances = set()
 
     @classmethod
-    def register_new(cls, name, func):
+    def _build_udf(cls, func, name=None):
```
eriknw, line 194: Hmm, probably no harm exposing this to the user, so no need to hide behind an underscore.

grblas/__init__.py
```diff
@@ -4,7 +4,7 @@
 _init_params = None
 _SPECIAL_ATTRS = ["lib", "ffi", "Matrix", "Vector", "Scalar",
                   "base", "exceptions", "matrix", "ops", "scalar", "vector"
-                  "unary", "binary", "monoid", "semiring"]
+                  "unary", "binary", "monoid", "semiring", "numpyops"]
```
jim22k, line 7: Rather than creating a new top-level module, I would rather have numpy functions registered as normal UDFs using nested names.

`grblas.numpyops.unary.sqrt` looks almost identical to `grblas.unary.numpy.sqrt`, but the latter doesn't need anonymous registration or a new top-level module.

test/test_op.py
```diff
@@ -58,26 +58,26 @@ def plus_one(x):
...
-                                    'FP32', 'FP64'}
+                                    'FP32', 'FP64', 'BOOL'}
     v = Vector.new_from_values([0, 1, 3], [1, 2, -4], dtype=dtypes.INT32)
     v << v.apply(unary.plus_one)
     result = Vector.new_from_values([0, 1, 3], [2, 3, -3], dtype=dtypes.INT32)
     assert v.isequal(result)
 
 
 def test_unaryop_udf_bool_result():
-    pytest.xfail('not sure why numba has trouble compiling this')
```
eriknw, line 69: Good news!  The hack to change BOOL to INT8 during computation appeared to work.  Well, at least this test passes now.

jim22k:
It sure would be nice to figure out ahead-of-time compiling with numba so we don't need to JIT compile the numpy functions every time we start up.

eriknw:
Yeah, that would be great (assuming the binary isn't too large).  One step at a time.

eriknw:
Whew!  Tests added.

UDFs over bools works for unary and binary operations, but not monoid or semiring.  Oh well.  Hopefully our hack to support bool unary and binary continues to work okay with few surprises.

Shall we merge?  I may eventually have `op.types` be a dict of `{input_type: output_type}` instead of just a set of input types.

PR #13, opened 2020-04-24
Fix bool monoids

eriknw:
This follows one of @seibert suggestions of casting the result back to `bool`.  I think the final solution makes sense, but it was an adventure of surprises to get there.  Score one for persistence.


eriknw:
I explicitly cast bool inputs to bool now as well.  Shall we merge?

This revealed that `np.reciprocal(False)` behaves differently for `numpy`(=0) and `numba` (=1).  File this under "won't fix" in `numba`.

PR #14, opened 2020-04-28
Parameterized udfs

eriknw:
When registering new unary or binary ops, use `parameterized=True` argument.  Monoids and semirings are implicitly parameterized if an input binaryop or monoid are parameterized.

This PR continued from https://github.com/jim22k/grblas/pull/13, so probably merge that one first.

### DIFF
--- a/grblas/vector.py
+++ b/grblas/vector.py
@@ -135,10 +135,7 @@ class Vector(GbContainer):
         if dup_op is None:
             dup_op = binary.plus
         else:
-            opclass = find_opclass(dup_op)
-            if opclass.startswith('Parameterized'):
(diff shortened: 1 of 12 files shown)


grblas/vector.py
```diff
@@ -233,6 +232,9 @@ def ewise_add(self, other, op=None, *, require_monoid=True):
         if op is None:
             op = monoid.plus
         opclass = find_opclass(op)
+        if opclass.startswith('Parameterized'):
```
jim22k, line 235: This code is duplicated in almost every function. Can we bury it inside `find_opclass` somehow?

eriknw: Done.  Also added tests for parameterized BinaryOp as accum.

PR #15, opened 2020-04-29
Add dtype and mask keyword arguments to `.dup`

eriknw:


grblas/scalar.py
```diff
@@ -50,11 +50,15 @@ def value(self, val):
...
         """Create a new Scalar by duplicating this one
         """
-        new_scalar = self.__class__.new(self.dtype)
-        new_scalar.value = self.value
+        if dtype is None:
+            new_scalar = self.__class__.new(self.dtype)
+            new_scalar.value = self.value
+        else:
+            new_scalar = self.__class__.new(dtype)
+            new_scalar.value = new_scalar.dtype.numba_type(self.value)
```
eriknw, line 61: An alternative to using `numba_type` on this line is:
```python
from . import ffi
new_scalar.value = ffi.cast(new_scalar.dtype.c_type, self.value)
```

PR #16, opened 2020-04-30
Scalar isequal isclose

eriknw:
Note that this continues from https://github.com/jim22k/grblas/pull/14 .  Sorry for the chain of dependent PRs.  I separated them since each may have particular things to discuss.

### DIFF
--- a/grblas/scalar.py
+++ b/grblas/scalar.py
@@ -1,3 +1,4 @@
+import grblas
 from .base import ffi, GbContainer
 from . import dtypes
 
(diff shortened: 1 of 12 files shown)


grblas/scalar.py
```diff
@@ -17,18 +18,68 @@ def __repr__(self):
...
+            check_dtype = False
+        if check_dtype and self.dtype != other.dtype:
+            return False
+        if self.is_empty or other.is_empty:
+            return self.is_empty is other.is_empty
+        # We can't yet call a UDF on a scalar, so lets convert to 1-d vector
+        left = grblas.vector.Vector.from_values([0], [self.value], size=1, dtype=self.dtype)
+        right = grblas.vector.Vector.from_values([0], [other.value], size=1, dtype=other.dtype)
+        matches = left.ewise_mult(right, grblas.binary.isclose(rel_tol, abs_tol)).new(dtype=bool)
+        return matches.reduce(grblas.monoid.land).value
```
jim22k, line 81: Is this actually faster than just calling `math.isclose(self.value, other.value, rel_tol=rel_tol, abs_tol=abs_tol)`?
I agree this is a good approach once GrB_Scalar exists, but for now, this seems overly complicated.

eriknw: No idea about performance.  Yes, it is complicated, but I trust it to be consistent.  I don't know if `math.isclose` will behave the same or not.

jim22k: Fair enough

eriknw: Oh yeah, I also want to expose the compiled numba function from UDFs.  Once we have this, I want to simplify this to something like
```python
grblas.binary.isclose.numba_func(self.value, other.value)
```

PR #18, opened 2020-05-04
Fix transposed matrix

eriknw:
Don't subclass TransposedMatrix from Matrix.  It's not really a GbContainer.  This led to surprising things to be possible.

Instead of blacklisting methods that don't work, choose to be positive and add the ones that do.  I hope this will make it harder to make mistakes in the future.

I took the liberty of moving `is_scalar` and `is_transposed` to private properties.

This PR continues from: https://github.com/jim22k/grblas/pull/15

### DIFF
--- a/grblas/base.py
+++ b/grblas/base.py
@@ -2,7 +2,7 @@ from . import lib, ffi
 from . import dtypes, ops, descriptor, unary
 from .exceptions import check_status
 from .ops import OpBase
-from .mask import Mask, ComplementedMask, StructuralMask, ValueMask
+from .mask import Mask, StructuralMask, ValueMask
 
 NULL = ffi.NULL
 
@@ -14,7 +14,7 @@ class Updater:
 
     def __getitem__(self, keys):
         # Occurs when user calls C(params)[index]; need something prepared to receive `<<` or `.update()`
-        if self.parent.is_scalar:
+        if self.parent._is_scalar:
             raise TypeError('Indexing not supported for Scalars')
         if type(keys) is IndexerResolver:
             resolved_indexes = keys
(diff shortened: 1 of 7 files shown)


grblas/base.py
```diff
@@ -49,20 +49,16 @@ def update(self, delayed):
...
         if not isinstance(gb_obj, ffi.CData):
             raise TypeError('Object passed to __init__ must be CData type')
         if not isinstance(dtype, dtypes.DataType):
             dtype = dtypes.lookup(dtype)
-
         self.gb_obj = gb_obj
         self.dtype = dtype
 
-    def __invert__(self):
-        return ComplementedMask(self)
```
jim22k, line 64: Removing this means a user cannot `(~M).V`. Probably not worth allowing.
If we are removing this, we should remove the ComplementedMask class.

Should we give a useful error message if someone tries to `mask=~M`? A message that masks need to indicate either `.V` or `.S`?

grblas/matrix.py
```diff
@@ -561,62 +556,67 @@ def _prep_for_assign(self, resolved_indexes, obj):
                     raise TypeError(f'Expected Matrix for assignment value; found {type(obj)}')
                 delayed = GbDelayed(lib.GrB_Matrix_assign,
                                     [obj.gb_obj[0], rows, rowsize, cols, colsize],
-                                    at=obj.is_transposed)
-
+                                    at=obj._is_transposed)
```
jim22k, line 559: `isinstance(obj, Matrix)` check above needs to also check for TransposedMatrix

eriknw: Good catch!  Fixed and added a test to exercise this.

grblas/base.py
eriknw: Good call on `ComplementedMask` (a coverage report would have hinted at this).  A useful error message here may not hurt.

PR #19, opened 2020-05-06
Create `TypedOp` classes for more flexibility.

eriknw:
This began by wanting to use the numba-compiled UDF for scalar `isclose`.
This resulted in paying down technical debt.
We no longer rely on weird global state, and some operations are more specific.

### DIFF
--- a/grblas/ops.py
+++ b/grblas/ops.py
@@ -13,6 +13,10 @@ from .exceptions import GrblasException, check_status
 UNKNOWN_OPCLASS = 'UnknownOpClass'
 
 
+def _normalize_type(type_):
+    return type_.name if isinstance(type_, dtypes.DataType) else type_
+
+
 class UdfParseError(GrblasException):
     pass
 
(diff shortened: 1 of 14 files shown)


grblas/ops.py
```diff
@@ -137,31 +200,29 @@ class OpBase:
...
-        self._specific_types[type_] = obj
+    def _add(self, op):
+        self._typed_ops[op.type] = op
+        self.types[op.type] = op.return_type
 
     def __delitem__(self, type_):
-        type_ = self._normalize_type(type_)
-        del self._specific_types[type_]
+        type_ = _normalize_type(type_)
+        del self._typed_ops[type_]
```
jim22k, line 221: Should `type_` also be removed from `self.types` here?

grblas/ops.py
```diff
@@ -590,39 +642,31 @@ def register_new(cls, name, monoid, binaryop):
...
-    return _return_type[gb_op]
+        opclass = gb_op.__class__.__name__
+    elif isinstance(gb_op, TypedOpBase):
+        opclass = gb_op.opclass
+    elif isinstance(gb_op, ParameterizedUdf):
+        gb_op = gb_op()  # Use default parameters of parameterized UDFs
+        gb_op, opclass = find_opclass(gb_op)
+    else:
+        opclass = UNKNOWN_OPCLASS
+    return gb_op, opclass
```
jim22k, line 669: Is this function used anywhere other than in the tests? If not, we should remove it and refactor the tests to use `get_typed_op` instead.

jim22k:
Overall, I like this change. Definitely helps unify various concepts floating around in different functions.

grblas/ops.py
eriknw: Good catch!

grblas/ops.py
eriknw: It's used one place: `grblas.base:GbContainer.__call__` to determine whether an argument in `*optional_mask_and_accum` is a binary op for accum.  I was hoping to remove `find_opclass` as well.

eriknw:
Thanks for looking at this.

Note that I changed the default op for `Matrix.reduce_rows` (and columns) to `binary.add`.  It no longer uses a different default for bool dtypes.  I prefer this for three reasons:
  1. I like the example in the signature so one clearly sees a binary op
  2. I don't want to guess what is means by `A.reduce_rows()`--does it count the true, use `lor`, or use `land`?
  3. It follows all other default operators such as `ewise_mult`, which don't have per-type defaults.

eriknw:
I'm glad you like this overall.  Please don't be shy about sharing small nits.  I've been changing a lot of things that didn't necessarily need changed to better match my sensibilities, but it's okay if you prefer things differently.

PR #31, opened 2020-07-10
Auto-import dtypes and others as appropriate.

eriknw:
Fixes https://github.com/jim22k/grblas/issues/30

Also added a test for sanity.


eriknw:
The added test should probably live elsewhere (i.e., a file that doesn't exist yet), but, meh.

PR #32, opened 2020-07-11
Fix reduce with scalar and accum (again)

eriknw:
Fixes https://github.com/jim22k/grblas/issues/28

My bad for not testing this better last time!  I got things backwards.


eriknw:
Merging Tuesday if no comment.

PR #33, opened 2020-07-11
This fixes a bug where temporaries were being garbage-collected too early

eriknw:
The fix is simple and pragmatic.
I still have half a mind to do some refactoring, but that may need to wait until we wrap another implementation!


eriknw:
This will likely get changed soon, but let's add this for now.

PR #35, opened 2020-07-12
Create better repr and repr_html.  Use pandas if available.

eriknw:
Notably, if a grblas Matrix or Vector is very large, we will grab only parts of it to put in pandas DataFrame. Hence, the pandas DataFrame doesn't need to match the dimensions the grblas object!  This is much better.

There is still more I would like to do with repr, but this seems like a good step! I still need to update the notebook and add some tests (preferably after some feedback).


eriknw:
Merging.  I've continued development on another branch.  I'll add tests there.

PR #37, opened 2020-07-27
Replace GbDelayed with typed expressions like MatrixExpression.

eriknw:
This PR touches lots of little things, and I move some things around such as the new `expr.py` file from `base.py`.
This is still a work in progress, and there are a couple rough spots to complete.
I plan to add nice repr to the `*Expression` objects.
I also plan to add more tests (gotta cover that code!).

### DIFF
--- a/grblas/__init__.py
+++ b/grblas/__init__.py
@@ -2,10 +2,12 @@ import importlib as _importlib
 from . import backends, mask  # noqa
 
 _init_params = None
-_SPECIAL_ATTRS = {"lib", "ffi", "Matrix", "Vector", "Scalar",
-                  "exceptions", "matrix", "ops", "scalar", "vector",
-                  "unary", "binary", "monoid", "semiring",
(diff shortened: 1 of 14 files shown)


eriknw:
@jim22k I'm going ahead and merging this PR, which includes many, many small things and a couple big things.  Feel free to leave comments on this PR.

I decided to hide the data (i.e., collapse details) for expression html reprs, but show data for everything else.  In an exploratory or learning setting, it's very nice to see the data without clicking.

The reprs for `Updater`, `AmbiguousAssignOrExtract`, and `IndexerResolver` are unchanged and could be improved in the future.  Btw, I moved these to `grblas.expr`, but this feels like the wrong name since it no longer has the Expression classes in it.  Maybe `grblas.core`?

We have much more thorough testing, and coverage (including branches) is very close to 100%.  We now test that all top-level items are importable, and that all the notebooks can execute.

PR #41, opened 2020-07-31
Move tests and use versioneer

eriknw:
Move tests into `grblas` and use versioneer.py to manage our version.


eriknw:
@jim22k this is ready for review, and I'd like your approval before merging.  This is mostly infrastructure things preparing for releasing to PyPI.  Do you have any objections to using versioneer.py?

eriknw:
An advantage of moving tests and distributing them with the package is it makes it easy to run the tests after installation no matter how it was installed.  Tests could be included as package data, but there's little harm IMHO in adding it as a module with `__init__`.

PR #50, opened 2020-08-18
Use SS:GraphBLAS header 3.3.3.  Oops.

eriknw:
Also, add a regression test that makes sures the header file versions match.


eriknw:
I'm very confused right now why this is failing.  I am able to create a fresh environment with a fresh repo, apply this patch, build, and run the tests.  @jim22k any ideas?  It's late, so I'm probably missing something silly.  I'll check again tomorrow.

eriknw:
Guess and try FTW!

PR #51, opened 2020-08-18
Catch dimension mismatch upon creation of the expression.

### LINKED ISSUE #40: Should we detect DimensionMismatch sooner?
For example, that `A.shape == B.shape` in `A.ewise_add(B)`, or that `A.ncols == B.nrows` in `A.mxm(B)`.

Now that delayed expressions are getting nicer reprs, it may be nice to detect these upon first creating the delayed expression.  Otherwise, it feels weird to look

The main downsides are a tiny performance hit and that our error message may be different from what people are used to from the underlying implementation.  A positive is that we'd have a consistent error message independent of implementation.

eriknw:
As opposed to calculating the expression with `.new()` or `<<`.

Fixes #40


eriknw:
I think this gives a nice experience, isn't too burdensome for us developers, and shouldn't slow things down too much.  I'm in favor.

PR #57, opened 2020-10-09
`to_values` returns numpy array

jim22k:
Make `to_values` return a tuple of numpy arrays.
Make `from_values` and `build` accept numpy arrays as input.
Add "size" option to build, allowing it to function as a more effective rebuild.

Significant speed improvement for large objects.
```python
import numpy as np
import grblas

ni = np.arange(100000)
nv = np.ones(100000)
li = ni.tolist()
lv = nv.tolist()

# This takes 4.3 ms with the previous code
vec = grblas.Vector.from_values(li, lv)

# This takes 0.25 ms with the new code
vec = grblas.Vector.from_values(ni, nv)
```
Exporting using `to_values()` has a similar speedup.

Note that `from_values` with list or tuple input actually slowed down with the new code. I put in additional checks for datatypes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ script:
     # Make sure all top-level imports work
     - ( for attr in Matrix Scalar Vector backends base binary descriptor dtypes exceptions expr ffi formatting init io lib mask matrix monoid ops scalar semiring tests unary vector ; do echo python -c \"from grblas import $attr\" ; if ! python -c "from grblas import $attr" ; then exit 1 ; fi ; done )
     # Make sure notebooks execute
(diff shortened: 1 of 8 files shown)


jim22k:
@eriknw , I couldn't figure out how to get numpy to raise an OverflowError when doing something like `np.array([0, 1, 2], dtype=bool). Same goes for attempting to find negative integers prior to calling `astype(np.uint64)`. Numpy seems to be happy to either overflow or coerce numbers, but won't raise an error. Am I missing something here?

eriknw:
LGTM

PR #58, opened 2020-11-17
Support subassign with expected syntax `C[I, J](Mask) << A`

eriknw:
Things became much simpler after introducing `Assigner`, which is the result of `C[I, J](Mask)` and `C(Mask)[I, J]` operations.  This was previously handled by `AmbiguousAssignOrExtract`.

This introduces SuiteSparse-specific functionality, and I note this in the code where appropriate.  The only recipes I needed to create were for assigning scalars for subassign to matrix row or column.

I flesh out some tests that were missing, such as `C(mask)[I, j] << v`.

I'm relying on CI to see if any notebooks need updated 🤞 

Oh, I should probably add something to the documentation (`README.md`).

### DIFF
--- a/grblas/matrix.py
+++ b/grblas/matrix.py
@@ -716,42 +716,95 @@ class Matrix(BaseType):
             if rowsize is None and colsize is not None:
                 # Row-only selection
                 row_index = rows
-                if is_submask:
-                    # SS, SuiteSparse-specific: subassign
-                    cfunc_name = "GrB_Row_subassign"
-                    expr_repr = "[{1}, [{3} cols]](%s) << {0.name}" % mask.name
+                if mask is not None and type(mask.mask) is Matrix:
+                    if is_submask:
+                        # C[i, J](M) << v
+                        raise TypeError("TODO")
+                    else:
+                        # C(M)[i, J] << v
+                        # Upcast v to a Matrix and use Matrix_assign
+                        rows = _CArray([rows.scalar.value])
+                        rowsize = _CScalar(1)
+                        new_value = Matrix.new(
+                            value.dtype, nrows=1, ncols=value.size, name=f"{value.name}_as_matrix"
+                        )
+                        new_value[0, :] = value
+                        delayed = MatrixExpression(
+                            method_name,
+                            "GrB_Matrix_assign",
+                            [new_value, rows, rowsize, cols, colsize],
+                            expr_repr="[[{2} rows], [{4} cols]] = {0.name}",
+                            nrows=self._nrows,
+                            ncols=self._ncols,
+                            dtype=self.dtype,
+                        )
                 else:
-                    cfunc_name = "GrB_Row_assign"
-                    expr_repr = "[{1}, [{3} cols]] = {0.name}"
-                delayed = MatrixExpression(
-                    method_name,
-                    cfunc_name,
-                    [value, row_index, cols, colsize],
-                    expr_repr=expr_repr,
-                    nrows=self._nrows,
-                    ncols=self._ncols,
-                    dtype=self.dtype,
-                )
+                    if is_submask:
+                        # C[i, J](m) << v
+                        # SS, SuiteSparse-specific: subassign
+                        cfunc_name = "GrB_Row_subassign"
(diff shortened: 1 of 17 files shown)


grblas/tests/test_matrix.py
```diff
@@ -399,6 +522,74 @@ def test_assign_column(A, v):
     assert C.isequal(result)
 
 
+def test_assign_row_scalar(A, v):
+    C = A.dup()
+    C[0, :](v.S) << v
+    D = A.dup()
+    D(v.S)[0, :] << v
```
jim22k, line 529: Are we not changing this behavior? I know this is how the GraphBLAS spec is written, but with subassign, this now feels wrong to be able to write this.

jim22k: After discussing, we will keep the original GraphBLAS spec behavior and also add a recipe for using a mask which matches the size of D as an alternate style for this call.

jim22k (approved):

grblas/matrix.py
```diff
@@ -777,21 +813,85 @@ def _prep_for_assign(self, resolved_indexes, value):
...
+                if colsize is None:
+                    cols = _CArray([cols.scalar.value])
+                    colsize = _CScalar(1)
+                if is_submask:
+                    # SS, SuiteSparse-specific: subassign
+                    cfunc_name = f"GrB_Matrix_subassign_{value.dtype}"
+                    expr_repr = "[[{2} rows], [{4} cols]](%s) = {0}" % mask.name
+                else:
+                    cfunc_name = f"GrB_Matrix_assign_{value.dtype}"
+                    expr_repr = "[[{2} rows], [{4} cols]] = {0}"
```
eriknw, line 885: How we handle scalar indices here results in some interesting behavior.  The code in this else statement handles scalar values on the RHS and Matrix or no mask.  Specifically:
```python
C[I, J](M) << c
C[I, j](M) << c
C[i, J](M) << c
C[i, j](M) << c
C[I, J] << c
C[I, j](<< c
C[i, J] << c
C[i, j] << c
C(M)[I, J] << c
C(M)[I, j](<< c
C(M)[i, J] << c
C(M)[i, j] << c
```
I don't think we should allow the following:
```python
C[I, j](M) << c
C[i, J](M) << c
C[i, j](M) << c
```
For assign, `C(mask)[indices] << value`, the mask must either match the matrix `C` or the indices (such as being a vector).

For subassign `C[indices](mask) << value`, the mask must match the indices.  There is no other option.

For both assign and subassign, the value must either match indices or be a scalar.

We do not currently treat a matrix with a dimension of length 1 as a vector, nor do we ever treat a vector as a matrix with a dimension of length 1.  Both of these behaviors are, of course, possible and safe, but are clearly "not yet".  Similarly, we could do the same for vector or matrix with one element and scalars.

jim22k: I agree with not allowing those options. Additionally, I wonder if we should disallow:
```python
C(M)[i, j] << c
```
Masks for a single element feel strange. But I'm not 100% against it either.

eriknw: It's cheap and low-complexity to support `C(M)[i, j] << c`, so I don't mind supporting mask (and accum, etc) here with a single element update with a scalar.

But, this does raise the option of `C[i, j].new(input_mask=Mask)`, which is probably fine-ish.  The assignment-equivalent of this is `c(input_mask=Mask) << C[i, j]`.  It's a little weird to use `input_mask` like this in a Scalar, but consistent.  `C[I, j].new(input_mask=Mask)` with vector result doesn't feel weird to me.

btw, partial update has been pushed (no `input_mask` yet).  There are a lot of error messages that need written to help guide people.

eriknw:
With `input_mask` argument, the following could be made to be equivalent for subassign:
```python
C[I, J](M) << A
```
```python
C(input_mask=M)[I, J] << A
```
This seems like perfectly reasonable behavior to expect to work.  For now, we only support `input_mask` for extract, and I prefer the first spelling for subassign.  There may be pressure to support more flexible `input_mask` in the future.

jim22k (approved):

PR #59, opened 2020-11-20
Add fast_import and fast_export methods to Vector and Matrix

jim22k:
These new functions bypass the recorder, mostly because I was just trying to get them working quickly, but also because the `**` double pointers would require changes to `_CArray` and I didn't want to mess with that right now.

### DIFF
--- a/grblas/matrix.py
+++ b/grblas/matrix.py
@@ -28,6 +28,8 @@ class Matrix(BaseType):
         self._nrows = None
         self._ncols = None
         super().__init__(gb_obj, dtype, name)
+        # Add ss extension methods
+        self.ss = Matrix.ss(self)
 
     def __del__(self):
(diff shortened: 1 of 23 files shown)


grblas/matrix.py
```diff
@@ -194,6 +194,106 @@ def to_values(self, *, dtype=None):
...
+         - x: values (ndarray of appropriate dtype)
+
+        To reimport the Matrix:
+        ```
+        pieces = A.fast_export()
+        A2 = Matrix.fast_import(**pieces)
+        ```
+
+        The underlying GraphBLAS object transfers ownership to numpy, disallowing further access.
+        The caller should delete or stop using the Matrix after calling `fast_export`.
```
eriknw, line 217: What is `self.gb_obj` after this call?  Should we set it to `ffi.NULL`?

PR #64, opened 2021-01-26
Changes necessary for Windows builds

jim22k:


grblas/backends/suitesparse/clean_header.py
```diff
@@ -14,7 +14,7 @@ def remove_directives(text):
     # There are a few cases of safe `#define` directives that we need to keep
     #  - #define FOO 12
     #  - #define BAR ...
-    safe_define = re.compile(r"^#define\s+\w+\s+(\d+|\.{3})")
+    safe_define = re.compile(r"^#define\s+\w+\s+(\d+|\.{3})\s*$")
```
eriknw, line 17: Why did you add the EOL back to this?  This was causing some definitions we need to be removed.

PR #65, opened 2021-01-27
Add `raw=` to export methods

eriknw:
For `raw=True`, we don't do any trimming or reshaping of the buffer we receive from SuiteSparse.  Otherwise, we try to make things as Python-friendly as possible.  We sometimes need to return extra items in the dict when `raw=True`.

Do we like the name `nvec` for hypersparse?


jim22k (approved):

eriknw:
I added a `my_matrix.ss.format` property that returns the string used in export (`"csr", "hypercsc", "bitmapr"`, etc).  Sound good?  Is this the best name?  Lingering things that would be nice to get in before releasing, but not strictly necessary:
1. `my_matrix.ss.format = 'CSR'`
1. Show the format in the reprs
1. Docstrings for import methods (\*sigh\*)
2. example notebook showing what's new (including other recent releases)

Note that transposed matrix does not have `.ss.format` (or export, etc), but it does have a pretty repr.  If we show the format in the repr (which would be very nice!), what format should we show for the transpose?

PR #66, opened 2021-01-27
Fixes to fully remove support for complex from grblas

jim22k:
Turns out removing support for complex numbers is tricky. Lots of assumptions baked into the code and tests about these being present.


jim22k:
I tested these changes on the Windows box and `conda build` works. The built package is able to pass all tests.

PR #67, opened 2021-02-18
Add `grblas.op` namespace to put all operations together.

eriknw:
Also, `grblas.op.numpy`, which remains lazy.

We now have `grblas.ops` and `grblas.op`.  Perhaps we should rename `grblas.ops`.

This was @jim22k's idea to help make `grblas` easier to use.


eriknw:
Some methods like `build` and `apply(op, right=1)` require a binaryop and don't accept monoids.  Having the monoids in `grblas.op` instead of binaryops may make using these methods more annoying.

Should we always convert monoids to binaryop when appropriate?  This should always be safe and possible to do, right?  This is similar to upgrading builtin binaryops to monoids when possible in ewise_add.

jim22k:
Given users can `grblas.binary.plus` and almost never need to access `grblas.ops`, lets rename it to `grblas.operators` or something like that.

And yes, auto-converting from monoid to binary or binary to monoid as required by a method seems okay to me.

jim22k (approved):

eriknw:
- Renamed to `grblas.ops` to `grblas.operator`.
- Allow monoids any place where binaryop is required.
- Allow binaryop where monoids are required if it is part of a monoid.
- Added attributes such as `binop.monoid`, `monoid.binaryop`, and `monoid.identity`.

All this made this PR much more substantial.

Question: can we not allow using a semiring where a binaryop or monoid is expected?  For example, with ewise_add.
```python
min_plus(A | B)  # bad.  does this do min or plus?
min_plus.binary(A | B)  # better
plus(A | B)  # best

min_plus(A & B)  # bad
min_plus.monoid(A & B)  # better
min(A & B)  # best
```

eriknw:
I should have said I plan to disallow using semiring where it grabs the binary/monoid from it such as ewise_add.  Let me know if you prefer otherwise.

grblas/operator.py
```diff
@@ -205,10 +250,11 @@ class OpBase:
     _initialized = False
     _module = None
 
-    def __init__(self, name):
+    def __init__(self, name, *, anonymous=False):
         self.name = name
         self._typed_ops = {}
         self.types = {}
+        self._anonymous = anonymous
```
eriknw, line 257: Should this be private or not?

eriknw:
Note that we no longer include Monoids in `grblas.op`.  BinaryOps are more general w.r.t. the types they can operate on.

eriknw:
@jim22k one more question for you.  When users register new operators, we assign them to e.g. `grblas.binary.my_udf` if they are not anonymous.  Should we also save it to e.g. `grblas.op.my_udf`?

grblas/tests/test_numpyops.py
```diff
@@ -177,19 +177,20 @@ def test_npmonoid():
...
+    #             [
+    #                 grblas.Vector.from_values(index, values1, dtype="FC64"),
+    #                 grblas.Vector.from_values(index, values2, dtype="FC64"),
+    #             ],
+    #             [
+    #                 np.array(values1, dtype=np.complex128),
+    #                 np.array(values2, dtype=np.complex128),
+    #             ],
+    #         ]
+    #     )
```
eriknw, line 193: Btw @jim22k complex monoids still don't work.  I apparently added a check in "operators.py" to avoid the segfault.

eriknw:
Well, this "small" PR became quite large.  Merging!

PR #68, opened 2021-02-19
Infix sugar

eriknw:
I think this is a decent start for adding the infix sugar as we discussed.  This is a WIP and needs tests and docs.

This adds things like the following
```python
C << op(A | B)  # ewise_plus
C << op(A & B)  # ewise_mult
C << op(A @ B)  # mxm
C << op(A)  # apply
C << op(A, 1)  # apply(right=1)
```
In the not-to-distant future, we'd like to spell select like `C << A[eq(A, 0)]`.

I'm lukewarm to the idea of adding comparison operators to do select like `C << A[A < 5]`.  However, doing apply like this **does not** work: `C << A < 5`.  One needs to do `C << (A < 5)`.  Definitely holding off on this for now.


setup.py
```diff
@@ -45,13 +49,14 @@
     long_description=long_description,
     long_description_content_type="text/markdown",
     author="Jim Kitchen and Erik Welch",
+    author_email="erik.n.welch@gmail.com",
```
eriknw, line 52: @jim22k would you like to include your email here?

eriknw:
@jim22k I think this is ready! 🎉

eriknw:
Functionally, anyway.  Still needs documented.  Lots of things should be documented better.  I think it's time to focus on documentation next.

setup.py
jim22k: sure

jim22k (approved):

jim22k:
Very nice work!

eriknw:
Thanks for reviewing @jim22k, I'm pretty excited to get this in!

PR #69, opened 2021-02-25
Add helper functions to take the "head" of vector or matrix objects.

eriknw:
This uses zero-copy export and import to try to be as efficient (including memory efficient) as possible, but it does modify `.gb_obj`, so it's a little funky.

I'm thinking of using this to improve the repr for sparse data.

This was pretty quick to pull together.  These aren't the most robust tests, but they're better than nothing.  Improving the repr is where the "real" work will be!


grblas/_ss_utils.py
```diff
@@ -0,0 +1,171 @@
...
+        elif fmt == "bitmap":
+            indices, vals = _head_indices_vector_bitmap(d["bitmap"], d["values"], d["size"], n)
+        elif fmt == "sparse":
+            indices = d["indices"][:n].copy()
+            vals = d["values"][:n].copy()
+        else:  # pragma: no cover
+            raise RuntimeError(f"Invalid format: {fmt}")
+    finally:
+        rebuilt = Vector.ss.import_any(take_ownership=True, name="", **d)
+        vector.gb_obj, rebuilt.gb_obj = rebuilt.gb_obj, vector.gb_obj
```
jim22k, line 48: Why reassign back to `rebuilt.gb_obj` here when `rebuilt` is thrown away?

jim22k (approved):
I really like this approach to taking ownership, grabbing a portion of the data, and giving the original data back. Nice trick!

Did you plan to add a method so this can be called as `myvec.ss.head(2)`?

grblas/_ss_utils.py
eriknw: It is necessary to replace `rebuilt.gb_obj` with *something*.  It would be sufficient to replace it with `NULL`.  Doing a swap here makes sure the old `gb_obj` gets cleaned up no matter what it is (right now it is `NULL`).  It also makes sure the original `rebuild.gb_obj` isn't cleaned up too early.  Perhaps I should add a comment in the code.

eriknw:
> Did you plan to add a method so this can be called as `myvec.ss.head(2)`?

I'm undecided.  I'm tempted to.  Probably.  What do you think?

I may also add `dtype=` keyword argument to `head` to make it match `to_values`.

I can't guarantee this method for any backend, so it would be inappropriate to add this functionality to `to_values` directly.

eriknw:
`myvec.ss.head(2)` added.  Also added `dtype=` keyword argument to make it behave more like `.to_values()`.

We don't have `A.T.ss` attribute yet, but we could add e.g. `A.T.ss.head()` (just like how there's `A.T.to_values()`) where we switch rows and cols.

grblas/formatting.py
```diff
@@ -202,18 +204,79 @@ def _get_matrix_dataframe(matrix, max_rows, min_rows, max_columns, *, mask=None)
                 column_offset,
                 mask=mask,
             )
+    if (
+        (mask is None or mask.structure)
+        and df.shape != matrix.shape
+        and min(matrix._nvals, max_rows if matrix._nvals <= max_rows else min_rows)
+        > 2 * df.count().sum()
```
eriknw, line 211: Note to self: this condition was chosen so I wouldn't need to rewrite any tests.  "Friendlier" conditions for when to switch to the sparse repr may exist.

grblas/formatting.py
```diff
@@ -202,18 +204,89 @@ def _get_matrix_dataframe(matrix, max_rows, min_rows, max_columns, *, mask=None)
...
+    ):
+        # The data is sparse and it's better to show in COO format.
+        # SS, SuiteSparse-specific: head
+        num_rows = matrix._nvals if matrix._nvals <= max_rows else min_rows
+        if matrix._is_transposed:
+            cols, rows, vals = matrix._matrix.ss.head(num_rows, sort=True)
```
eriknw, line 217: Note to self: sorting here for a better repr result.  If we find this is too slow for large objects, we can update it so we only sort for small objects (but we need to figure out where the threshold may be).

PR #72, opened 2021-03-11
Serialize Scalar, Vector, and Matrix

eriknw:
- Add `__slots__` to all the things
  - it simplifies serializing Scalar, Vector, and Matrix (don't need to worry about `__getstate__` and `__setstate__`)
  - we may want to serialize other things eventually too (when there is need)
  - slightly faster and less memory usage
  -  behaves similar to if we wrapped things in Cython
- Move SuiteSparse-specific functions and things to `grblas._ss` namespace.
  - We can rename to `grblas.ss` once this has useful things for users.
- Give `ParameterizedBinaryOp` a `.monoid` attribute just like other binaryops.

Note that pickling Vector and Matrix objects uses export/import from SuiteSparse.  I haven't been very strict about checking the backend.  It would help to have a non-SuiteSparse backend to test.


eriknw:
I think this is ready to go in.

jim22k (approved):

PR #79, opened 2021-03-30
Don't allow expressions with `v[w]`; we don't support boolean maskig like numpy

eriknw:
Instead, use `mask=` keyword in the appropriate operation.
Also, for sanity, don't allow Vector, Matrix, or TransposedMatrix to be converted to numpy arrays.
Heads up: Some array creation semantics are changing in numpy 1.20.

### DIFF
--- a/grblas/base.py
+++ b/grblas/base.py
@@ -406,6 +406,15 @@ class BaseType:
     _expect_type = _expect_type
     _expect_op = _expect_op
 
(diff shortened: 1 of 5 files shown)


eriknw:
I don't know of any other library defining `__array__` or `__array_struct__` to prevent from being converted to a numpy array, but this seems fine to me.  Merging.

PR #84, opened 2021-04-23
Changes needed to depend on `python-suitesparse-graphblas`.

eriknw:
This currently works with the PR: GraphBLAS/python-suitesparse-graphblas#9

Of course, this PR won't pass until we have package on conda-forge for `suitesparse-graphblas`


eriknw:
I think this is ready, but we may want to wait to merge until `python-suitesparse-graphblas` is on conda-forge and has wheels for all platforms.

eriknw:
I'm going for it!  Merging and releasing to PyPI.

PR #87, opened 2021-05-05
Add Aggregators (WIP)!

eriknw:
Aggregators are pretty far along, but I still need to do a few things:
- Handle dtypes better
- Give better names to objects so the recorder looks better
- Add a few more tests
- `first` and `last` for `Matrix.reduce_scalar`
- General clean up

Additional changes:
- Allow Scalars to be used in indexing
- Allow Scalars to be used to construct numpy arrays
- Fix SuiteSparse bug in Matrix apply1st and apply2nd
- Fix dtypes.unity for complex numbers (all pairs now match numpy behavior)

XREF: https://github.com/DrTimothyAldenDavis/GraphBLAS/issues/33


eriknw:
Continued in https://github.com/metagraph-dev/grblas/pull/104

PR #88, opened 2021-05-14
Update to 5.0.3.  Includes changes that were part of the aggregator branch.

eriknw:
- Allow Scalars to be used in indexing
- Allow Scalars to be used to construct numpy arrays
- Fix SuiteSparse bug in Matrix apply1st and apply2nd (but now fixed in SS 5.0.3)
- Fix dtypes.unity for complex numbers (all pairs now match numpy behavior)
- Add sane coercions to operators (such as `exp(1)`)


eriknw:
Merging and releasing!

PR #89, opened 2021-07-01
WIP: add prefix scan methods to do e.g. cumsum and cumprod.

eriknw:
Added:
- `my_matrix.ss.scan_rows(optional_monoid)`
- `my_matrix.ss.scan_columns(optional_monoid)`
- `my_vector.ss.scan(optional_monoid)`

We perform the scan with a series of matrix multiplies a la the Blelloch method.  See:
https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda

I bet there are faster ways to do this and that the current approach could be further optimized.  Nevertheless, I think this is a nice and useful API, so I'd like to add it to help push the boundary.  I think we could find many use cases for this.

We could turn this into an expression so that masks, accumulators, etc. could be used, but I haven't worried about such semantics yet (and don't know of a use case).  We could also further extend this perform a transpose, such as `my_matrix.T.ss.scan_rows()`.

TODO:
- [x] document
- [x] use iso matrices from GraphBLAS 5.1 for `S` matrices
- [x] test use of different monoids
- [x] improve and test `get_semiring`
- [ ] benchmark and optimize?


eriknw:
All done!

I merged the aggregator PR into this one, but it turns out we don't need it.  Oops.  I'll finish aggregators soon, so no big deal.

I'm eager to benchmark this, but I don't have the time yet, and not benchmarking shouldn't prevent being merged.

This should be fast enough and memory-efficient enough for all use cases 🤞

PR #90, opened 2021-07-02
Update to SS:GraphBLAS 5.1.2

eriknw:


eriknw:
Btw, we don't yet support GxB functions that use `GxB_Scalar` objects.  It's probably time to change that.  For example, we don't yet have an easy way to call `GxB_Matrix_build_Scalar` to create iso-values matrices.

Merging!

PR #95, opened 2021-07-12
Add `.new()` method to infix expressions.

eriknw:
In particular, I have found myself wanting to do `(A @ B).new()`.


eriknw:
Merging!  If we don't like the defaults operators here, then we should update the defaults in their respective methods.

PR #98, opened 2021-07-15
WIP: add methods to Expression objects that automatically compute `.new()`

eriknw:
@jim22k I'll want your approval before merging this.

The idea is to automatically compute `.new()` when the context is obvious.  In practice, typing `.new()` all the time is kind of a hindrance (even though it's explicit).  Often times, the intent is clear with just the expression.

I still need to update all the type-checking and such, but I think this will be mostly straightforward.

**Question:** should we automatically compute the result and show it when displaying the html repr of expressions?  I'm inclined to say no in case the computed result is huge.


grblas/_automethods.py
```diff
@@ -0,0 +1,315 @@
...
+print("    # These raise exceptions")
+for name in sorted(common_raises | vector_matrix_raises):
+    print(f"    {name} = wrapdoc(Matrix.{name})(Matrix.{name})")
+
+```
+"""
+
+
+def _get_value(self):
+    # A config to control auto-compute would go here
```
eriknw, line 120: Note to self: this is where a config to control auto-compute would go

grblas/tests/test_matrix.py
```diff
@@ -2103,3 +2105,86 @@ def test_concat(A):
...
+        assert expr.vxm(A).new(mask=expr.V).isequal(expected.vxm(A).new(mask=expected.V))
+
+
+def test_expr_is_like_matrix(A):
+    attrs = {attr for attr, val in inspect.getmembers(A)}
+    expr_attrs = {attr for attr, val in inspect.getmembers(binary.times(A & A))}
+    infix_attrs = {attr for attr, val in inspect.getmembers(A & A)}
+    transposed_attrs = {attr for attr, val in inspect.getmembers(A.T)}
+    # Should we make any of these raise informative errors?
+    expected = {
```
eriknw, line 2159: This is a good place to look for whether expressions "look like" the original object.

grblas/tests/test_scalar.py
```diff
@@ -261,3 +262,28 @@ def test_neg():
...
             assert (-empty).value is None
+
+
+def test_expr_is_like_vector(s):
+    v = grblas.Vector.from_values([1], [2])
+    attrs = {attr for attr, val in inspect.getmembers(s)}
+    expr_attrs = {attr for attr, val in inspect.getmembers(v.inner(v))}
+    infix_attrs = {attr for attr, val in inspect.getmembers(v @ v)}
+    # Should we make any of these raise informative errors?
+    expected = {
```
eriknw, line 273: And here

grblas/tests/test_vector.py
```diff
@@ -1056,3 +1061,100 @@ def test_outer(v):
...
+        assert complex(s1) == complex(s2)
+        assert_array_equal(np.array([s1]), np.array([s2]))
+
+
+def test_expr_is_like_vector(v):
+    attrs = {attr for attr, val in inspect.getmembers(v)}
+    expr_attrs = {attr for attr, val in inspect.getmembers(binary.times(v & v))}
+    infix_attrs = {attr for attr, val in inspect.getmembers(v & v)}
+    # Should we make any of these raise informative errors?
+    expected = {
```
eriknw, line 1135: And here

grblas/tests/test_matrix.py
```diff
@@ -2103,3 +2105,86 @@ def test_concat(A):
...
+        assert expr.vxm(A).new(mask=expr.S).isequal(expected.vxm(A).new(mask=expected.S))
+        assert expr.vxm(A).new(mask=expr.V).isequal(expected.vxm(A).new(mask=expected.V))
+
+
+def test_expr_is_like_matrix(A):
+    attrs = {attr for attr, val in inspect.getmembers(A)}
+    expr_attrs = {attr for attr, val in inspect.getmembers(binary.times(A & A))}
+    infix_attrs = {attr for attr, val in inspect.getmembers(A & A)}
+    transposed_attrs = {attr for attr, val in inspect.getmembers(A.T)}
+    # Should we make any of these raise informative errors?
```
eriknw, line 2158: This is a good question.  We could give more informative error messages.  Would this be helpful?

eriknw (commented):
I think this is ready!  Hopefully coveralls can provide coverage info (it's been having issues).

eriknw:
I added `.ss` accessor to the expressions.  This poses some risk, because some of these methods can mutate the underlying object.  Otherwise, we treat the cached, auto-computed value as read-only.

eriknw:
Merging!

This doesn't handle `AmbiguousAssignOrExtract` yet, so one can't do `v[:3] = v[1:4]`.  However, it would be nice to improve `AmbiguousAssignOrExtract` go have a better repr, so maybe we can make it auto-compute as appropriate like other expressions at the same time.

One other thought: one needs to do `.new()` to view the computed expression.  What if we had something like `expr.view` that could be used to show the REPR?  Punting on this for now.

PR #102, opened 2021-08-13
Add "coo" as import/export option for SuiteSparse.

eriknw:
Not only is it nice to have this alongside other import/export options, it lets us expose additional functionality that SuiteSparse provides:

- is_iso, import/export values array with a single element
- sorting on output

Also, added `x.ss.build_scalar` for GxB_Matrix_build_Scalar and GxB_Vector_build_Scalar.  Finally, accept a scalar for the values in `from_values`.

Note that I changed `sorted_index` to `sorted_rows` and `sorted_cols` for Matrix import/exports.  This lets us distinguish whether COO indices are sorted row-first or column-first.  Vector import/export still uses `sorted_index`.

### DIFF
--- a/grblas/_ss/matrix.py
+++ b/grblas/_ss/matrix.py
@@ -688,7 +688,7 @@ class ss:
                 scalar.value = nvals
                 call(
                     f"GrB_Matrix_extractTuples_{parent.dtype.name}",
-                    [rows, columns, ffi.NULL, _Pointer(scalar), parent],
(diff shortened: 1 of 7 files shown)


eriknw:
This may not be the final spelling.  In particular, in all the other import functions, one indicates whether the result should be row or column oriented.  In this case, should we have "coor" and "cooc" ("cor" and "coc")?  In this case, "sorted_rows" and "sorted_cols" could be used by both methods to indicate to what degree things are sorted.

PR #104, opened 2021-09-07
Handle aggregator dtypes in a more systematic and sensible way

eriknw:
This is a continuation of https://github.com/metagraph-dev/grblas/pull/87


eriknw:
:tada:

Btw, `argmini` and similar were dropped in the end.  `argmin` does the expected thing, but doesn't work with `matrix.reduce_scalar`.

PR #107, opened 2021-09-10
Add `matrix.ss.flatten_rows` to flatten a Matrix to a Vector row-wise.

eriknw:
Also `flatten_columns` of course!
We could also spell this as `matrix.ss.flatten()`, but then what do we use as arguments?
TODO: add the inverse operation to go from Vector to Matrix

### DIFF
--- a/grblas/_ss/matrix.py
+++ b/grblas/_ss/matrix.py
@@ -23,6 +23,7 @@ from ..utils import (
 )
 from .prefix_scan import prefix_scan
 from .scalar import gxb_scalar
+from .utils import get_order
 
 ffi_new = ffi.new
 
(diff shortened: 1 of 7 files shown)


eriknw:
The NumPy way to spell this is:
- `flatten(order='C')` for `flatten_rows`
  - inverse: `reshape((nrows, ncols), order='C')`
- `flatten(order='F')` for `flatten_columns`
  - inverse: `reshape((nrows, ncols), order='F')`

How should we spell the inverse? `vector.ss.reshape_rowwise(nrows, ncols)`?

`np.ravel` also flattens an array, but tries to avoid copying.

eriknw:
Merging!

This PR also reworks importing and exporting COO format.  I think it's more sensible.  There is now `coor` and `cooc` for COO that is sorted row-wise and column-wise respectively.

I'm still considering whether `v.ss.reshape` may also accept a tuple for the shape.  For example, NumPy arrays support both `a.reshape((5, 3))` and `a.reshape(5, 3)`.  The former is natural for doing e.g. `a.reshape(b.shape)`.  Of course, `a.reshape(*b.shape)` isn't difficult to do either.  Actually, yeah, maybe we should match NumPy.

PR #108, opened 2021-09-12
WIP: Add config option whether or not to auto-compute values.

eriknw:
This uses donfig.


eriknw:
Ready!

@jim22k note that disabling autocompute `grblas.config.set(autocompute=False)` also disables `scalar_expr.value`.  Previously, we auto-computed scalar expressions for things like this.

In general, I like autocompute when typing in an interactive session, but I prefer to disable it when writing library code or production code.

PR #131, opened 2021-12-01
Import grblas faster by delaying creation of new ops.

### LINKED ISSUE #130: compared to other packages `from grblas import ...` is 10 times slower
Hi @eriknw 

This may be a minor issue which I noticed a while back: `from grblas import ...` is unexpectedly slow.  Pls take a look at this [notebook](https://github.com/ParticularMiner/grblas/blob/import/notebooks/import_times.ipynb) and let me know if my concern is valid.

eriknw:
Fixes #130.  CC @ParticularMiner

Now the import time of grblas is dominated by importing pandas, numba, and numpy.  For me, the import time is about 5x faster (hooray!).

This was a bit more complicated than I expected.  I'll give this another pass tomorrow.

### DIFF
--- a/grblas/_agg.py
+++ b/grblas/_agg.py
@@ -48,6 +48,7 @@ class Aggregator:
         types=None,
     ):
         self.name = name
+        self._initval_orig = initval
         self._initval = False if initval is None else initval
         self._initdtype = lookup_dtype(type(self._initval))
         self._monoid = monoid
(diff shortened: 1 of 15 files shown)


ParticularMiner:
@eriknw 

A considerable improvement!  For me it is 3 to 7 times faster than before after merging your changes.

PR #134, opened 2021-12-12
Fix e.g. `import grblas.matrix` when done before `grblas.init` is called

eriknw:
This bug was breaking deserialization of Matrix objects.  `sys.meta_path` ftw!


eriknw:
We may want to puzzle out how to remove the import cycle that can sometimes result.  Although this is a lightweight use of `sys.meta_path`, most people aren't familiar with or comfortable with the particulars of Python's import machinery.  But, the current solution works splendidly!

PR #137, opened 2021-12-16
proposed empty_like() method for Scalar, Vector and Matrix

ParticularMiner:
Hi @eriknw

Here's a proposal for a small but useful addition to `grblas` which could go a long way to shorten code in `dask-grblas`, specifically the masked `matmul` code:

* add a method `empty_like()` to classes `Scalar`, `Vector`, and `Matrix`.

For example, in the chunk-function `_matmul2_masked` in `dask-grblas`:

```python
def _matmul2_masked(op, dtype, at, bt, mask, a, b, computing_meta=None):
    left = _transpose_if(a, at)
    right = _transpose_if(b, bt)
    mask = mask_type(mask.value)
    C = mask.mask.empty_like(dtype)
    C(mask=mask) << op(left @ right)
    return C
```
the parameter `mask` has a `mask` property that can be either a `Scalar`, `Vector`, or `Matrix`.  So having a method like `empty_like()` avoids having to code separate branches for each type.

Having said that, please feel free to direct me to a better alternative if you have one.

### DIFF
--- a/grblas/matrix.py
+++ b/grblas/matrix.py
@@ -325,6 +325,12 @@ class Matrix(BaseType):
         rv._ncols = ncols.scalar.value
         return rv
 
+    def empty_like(self, dtype, *, name=None):
+        """
+        Create a new empty Matrix from the given type and the same shape as this one
+        """
+        return Matrix.new(dtype, nrows=self.nrows, ncols=self.ncols, name=name)
+
     @classmethod
     def from_values(
         cls,
--- a/grblas/scalar.py
+++ b/grblas/scalar.py
@@ -280,6 +280,12 @@ class ScalarExpression(BaseExpression):
     def new(self, dtype=None, *, name=None):
         return super().new(dtype, name=name)
 
+    def empty_like(self, dtype, *, name=None):
+        """
+        Create a new empty Scalar from the given type like this one
+        """
+        return self.new(dtype, name=name)
+
     dup = new
 
     def __repr__(self):
--- a/grblas/vector.py
+++ b/grblas/vector.py
@@ -286,6 +286,12 @@ class Vector(BaseType):
         rv._size = size.scalar.value
         return rv
 
+    def empty_like(self, dtype, *, name=None):
+        """
+        Create a new empty Vector from the given type and the same size as this one
+        """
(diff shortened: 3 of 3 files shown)


ParticularMiner:
@eriknw 

> Could you make the PR against https://github.com/metagraph-dev/grblas?

Done.

> I like the idea behind this, and I think we could add more convenience functions like this. I wonder whether we should try to better match NumPy's `empty_like`, which is a top-level function, `np.empty_like`, that takes optional `dtype` (it also takes optional `shape` to let you override it).

Interface standardization across popular packages like `numpy` is more often than not a plus as far as user-experience is concerned.  So I’m all for it!

eriknw:
Your specific example
```python
    C = mask.mask.empty_like(dtype)
    C(mask=mask) << op(left @ right)
```
can also be done as
```python
    C = op(left @ right).new(dtype, mask=mask.mask)
```
which I would probably prefer in this case.

I think there may be an opportunity for `empty_like`, especially if it retains the flags from the original object such as format (although we don't yet expose full control of these in an easy manner).

ParticularMiner:
@eriknw 

In fact, your suggestion is even better as it makes `dask-grblas` code even shorter by obviating the need for separate chunk-functions for masks.  Perfect!

jim22k:
I propose that we re-use `M.new()` for this functionality once the deprecation cycle has finished. Move it from a classmethod to an instancemethod and have all the defaults (size/nrows/ncols, dtype, transpose flag, etc) come from the instance. It is essentially the same as `M.dup().clear()`, but more efficient.

And I do like that the same method name can be used across Matrix, Vector, Scalar to avoid needing to handle `.ndims` for this case.

eriknw:
I don't like using `.new()` for this.  I expect `.new()` to evaluate an expression, not create an _empty_ object.

What about adding an optional `clear=False` keyword to `dup`?  `build` also has `clear=False` keyword.

eriknw:
Closing in favor of #275.

`empty_like` is a pretty good suggestion (and it has a good name), so thanks for raising this @ParticularMiner!  I hope users will be able to find and use `B = A.dup(clear=True)` instead of `B = A.empty_like()`.

PR #138, opened 2021-12-19
Make `gb.ss.concat` work with vectors!

eriknw:
When creating a Matrix, a Vector is handled as an Nx1 Matrix.

I still need to update the docstring for usage.
I would also like to add `v.ss.split` for Vectors.

For example:
```python
>>> gb.ss.concat([v, v])  # Vector, size=2*N
>>> gb.ss.concat([[v], [v]])  # Matrix, nrows=2*N, ncols=1
>>> gb.ss.concat([[v, v]])  # Matrix, nrows=N, ncols=2
```

### DIFF
--- a/grblas/_ss/matrix.py
+++ b/grblas/_ss/matrix.py
@@ -193,15 +193,18 @@ def normalize_chunks(chunks, shape):
     if isinstance(chunks, (list, tuple)):
         pass
     elif isinstance(chunks, Number):
-        chunks = (chunks, chunks)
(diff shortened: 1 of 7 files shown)


eriknw:
Now that Matrix objects have `._parent` attribute that are responsible for freeing memory, we could make `from_pygraphblas` not invalidate the pygraphblas Matrix.  They could both remain valid while using the same underlying GraphBLAS object.

PR #140, opened 2022-01-04
A few changes to be more friendly to `dask-grblas`.

eriknw:
Most notably, `dask_grblas` is now able to automatically copy and convert testsfrom `grblas`.  This is important so that `dask-grblas` is able to better match the `grblas` API *and* to stay up to date more easily.  It is the burden of `dask-grblas` developers to ensure `grblas` tests may be copied over.

This entails adding a superfluous (for `grblas`) function `compute` in a few tests. It also includes separating imports of `Matrix`, `Vector`, and `Scalar` so these may be easily imported from `dask_grblas` instead of `grblas`. Finally, I changed some argument names from `delayed` to `expr`, since `delayed` means something else in `dask_grblas`.

### DIFF
--- a/grblas/base.py
+++ b/grblas/base.py
@@ -299,38 +299,38 @@ class BaseType:
             "Perhaps use .nvals attribute instead."
         )
 
-    def __lshift__(self, delayed):
-        return self._update(delayed)
+    def __lshift__(self, expr):
+        return self._update(expr)
 
-    def update(self, delayed):
+    def update(self, expr):
         """
         Convenience function when no output arguments (mask, accum, replace) are used
         """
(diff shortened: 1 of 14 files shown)


eriknw:
btw, the code that copies the tests is pretty low tech: https://github.com/metagraph-dev/dask-grblas/blob/4b5ac2a68e80eaa25fe89e78680a93f0738f201a/tests/from_grblas2/copytests.py

jim22k (approved):
The need for `compute` could be sticky issue for future contributors. Documentation should be added to explain when and why it is needed. But given the close connection between grblas and dask-grblas, this is fine for now.

PR #144, opened 2022-01-12
Add `.ss.compactify` to Vector and Matrix objects

eriknw:
Vectors are still TODO.  We also need tests.

We had previously spec'ed out having `sort=` argument to compactify.
We can still add this, but I don't know if it's really necessary.


eriknw:
Probably good enough.  Merging.  It may be nice to actually show examples.

PR #159, opened 2022-02-02
Ewise-union

### LINKED ISSUE #148: IDEA: allow strings to be used in place of operators `A.mxm(B, 'min.+')`
We now have `op.from_string`, `unary.from_string`, etc., functions that are used to get an operator from a string.  It would be straightforward to allow strings to be used wherever operators are used, such as:
- `A.mxm(B, "min.+")`
- `A.apply("+", 1)`
- `C("+") << A`
  - we'll need to distinguish "replace" string from operator strings

This is sugar, and, hence, pretty low priority.  Thoughts?

### LINKED ISSUE #154: ENH: use GxB_eWiseUnion when using `x - y` infix
As has been previously discussed, we would like to support `x - y` between Vector and Matrix objects that behaves like `eWiseAdd`, but with a default of `0` for the left value if it is not present.  This is what `GxB_Vector_eWiseUnion` and `GxB_Matrix_eWiseUnion` methods do.

This is currently SuiteSparse:GraphBLAS-specific behavior, but it would be possible to write a recipe that also does this if we ever need to support a non-SuiteSparse implementation.

eriknw:
Fixes #154 
Fixes #148 
This also added a `_GrbScalar` wrapper for making C calls.
Docstrings and tests still needed.
For now, the implementation for `ewise_union` is SuiteSparse-specific.  We could make a recipe when needed.
Also, hooray for #157, which made adding `ewise_union` method that much easier!

I'm not 100% sold on the API
```python
def ewise_union(self, other, op, left_default, right_default):
```
@jim22k what do you think?


eriknw:
Pop quiz: what do y'all think `x - y` will do (or should do) when `x` and `y` both have unsigned integer dtypes?  Similarly, what about with `-x`?

ParticularMiner:
I'm assuming you're hinting at overflows?  If so, then perhaps we should follow [`numpy`'s style](https://numpy.org/doc/stable/reference/generated/numpy.sum.html), which is to assume modular arithmetic and in so doing ignore overflows.

eriknw:
Yup, that's exactly right.  The result stays an unsigned integer.  This follows numpy rules.

jim22k:
> Fixes #154 Fixes #148 This also added a `_GrbScalar` wrapper for making C calls. Docstrings and tests still needed. For now, the implementation for `ewise_union` is SuiteSparse-specific. We could make a recipe when needed. Also, hooray for #157, which made adding `ewise_union` method that much easier!
> 
> I'm not 100% sold on the API
> 
> ```python
> def ewise_union(self, other, op, left_default, right_default):
> ```
> 
> @jim22k what do you think?

I like the ordering, but the names `left_default` and `right_default` leave a lot to be desired. Maybe we should think of them as fill values rather than defaults. What do you think of `lfill` and `rfill`? I'm all about brevity if we don't lose too much clarity in the names.

eriknw:
I don't like `rfill` and `lfill`.

How about `default`, which may be a single scalar or an iterable with two items?  I suspect it may be common to use the same value for both the left and the right.

btw, SuiteSparse:GraphBLAS calls these `alpha` and `beta`.

eriknw:
An even funkier spelling that is both concise and convenient could be:
```python
def ewise_union(self, other, op, *defaults):
```
where `defaults` could be length 1 or 2 (or 0 if we ever want a default value).

eriknw:
Okay, I think this (and all my current PRs) is ready pending a final decision on function signature.

eriknw:
Merging as is.  @jim22k we can continue discussing the function signature and make changes before releasing if necessary.

PR #162, opened 2022-02-03
Allow reduce to scalar to result in empty scalar

### LINKED ISSUE #160: Allow reduce to scalar to result in empty scalar
It would be nice to have an option to allow reduce to scalar to result in an empty scalar rather than the monoid identity.  By adding this to the API, we can make some of the aggregators in `dask-grblas` much faster by not requiring computing `nvals`.

@jim22k what do you think?  What's a good keyword?  `C.reduce_scalar(allow_empty=True)`?  Default `True` or `False`?

Perhaps we should bubble this up to SuiteSparse:GraphBLAS to e.g. allow the behavior to be toggled via a descriptor.

CC @ParticularMiner

eriknw:
Fixes #160

This should be able to speed up some aggregators in `dask-grblas`.

### DIFF
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -49,7 +49,7 @@ jobs:
         run: |
           source "$CONDA/etc/profile.d/conda.sh"
           conda activate grblas
-          conda install -c conda-forge pandas numba scipy networkx cffi
+          conda install -c conda-forge pandas numba scipy networkx cffi donfig
           if [[ ${{ matrix.sourcetype }} == "wheel" ]]; then
               pip install suitesparse-graphblas
           else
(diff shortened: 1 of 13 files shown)


eriknw:
Okay, so reducing to `GrB_Scalar` _will_ result in an empty Scalar if the Vector or Matrix is empty.  I guess I didn't read carefully enough.

ParticularMiner:
> Okay, so reducing to `GrB_Scalar` _will_ result in an empty Scalar if the Vector or Matrix is empty. I guess I didn't read carefully enough.

So this means there is no need to check for emptiness of `grblas` collections?

eriknw:
> So this means there is no need to check for emptiness of grblas collections?

That is the goal of this PR: to avoid checking `nvals` like we were doing before when reducing scalar.

I still like the current API that has the option of specifying whether the result can be empty or should use the monoid identity.  One often wants the identity.  Sometimes one doesn't.  So, it may make sense for `grblas.Scalar` to either be backed by a C scalar (as is currently done) or a `GrB_Scalar`.  Then we can choose how we want reduce to scalar to behave based on the underlying scalar type.

eriknw:
Merging!  I enabled coverage report to work even when some tasks fail.

There may be some lingering tasks:
- Check usage of reductions in notebooks
- Coverage

PR #163, opened 2022-02-07
Allow Scalar objects to be backed by GrB_Scalar.

### LINKED ISSUE #151: ENH: Update to use GrB_Scalar
We should look into updating our `Scalar` object to use `GrB_Scalar`.

One surprising side-effect of doing so would be to cast a scalar into a length 1 Vector or 1x1 Matrix in SuiteSparse:GraphBLAS.  This means it would be easy to do e.g. `s.ewise_add(t, op.plus)`.

eriknw:
I got distracted this weekend and took a stab at adding `GrB_Scalar`.  This is a lot of code change for what amounts to almost no change for the Python user.

Fixes #151.

Scalars may now be C scalars or GrB_Scalar.  We currently use `is_cscalar=` keyword.

There is a worrisome segfault showing up now.  Needs investigated.  This can be caused with:
```python
from grblas import Scalar

Scalar.from_value(1j, dtype="FC64")
```
Right now, I think this PR gets us 95% towards property supporting GrB_Scalar.
I want more thorough testing and to bring code coverage back up.

Feedback welcome.  How awkward is it to have two different modes for Scalars? `Scalar.from_value(2, is_cscalar=True)` and `Scalar.from_value(2, is_cscalar=False)`?  Do we like the name `is_cscalar` as the keyword and for Scalars to have attributes `s.is_cscalar` and `s.is_grbscalar`?


grblas/tests/test_scalar.py
```diff
@@ -322,3 +323,37 @@ def test_ndim(s):
     v = Vector.from_values([1], [2])
     assert v.inner(v).ndim == 0
     assert (v @ v).ndim == 0
+
+
+@pytest.mark.skipif("not dtypes._supports_complex")
+# @pytest.mark.parametrize("dtype", ["FC32", "FC64"])  # This segfaults
+@pytest.mark.parametrize("dtype", ["FC64", "FC32"])
```
eriknw, line 330: Note that the order we test the dtypes determines whether we segfault or not (on line 337 below).

grblas/tests/test_scalar.py
```diff
@@ -252,9 +249,11 @@ def test_scalar_to_numpy(s):
 
 
 def test_neg():
-    for dtype in vars(dtypes).values():
-        if not isinstance(dtype, dtypes.DataType):
-            continue
+    for dtype in sorted(
+        (dtype for dtype in vars(dtypes).values() if isinstance(dtype, dtypes.DataType)),
+        key=lambda x: x.name,
+        reverse=True,  # XXX: segfault when False!!!
```
eriknw, line 255: Note that the order we test the dtypes determines whether we segfault or not.

eriknw:
###  _"with this one ridiculously easy trick, you can test using both C scalars and GrB scalars as the default scalar types!"_
Heh, the last two commits are pretty fun.  Note that this works, because I chose to use `GrB_Scalar`s everywhere by default.

eriknw:
Alright, I think this PR is more-or-less ready!  @jim22k and/or @ParticularMiner I'd be happy to do a high-level walkthrough if you're interested.

There are a couple notable differences between C-backed `Scalars` and GrB-backed `Scalars`:
- Lossy coercion (e.g., `int_scalar.value = 12.5`)
  - C scalar raises (b/c `cffi` raises for us)
  - `GrB_Scalar` coerces
  - We could change either of these behaviors, but changing C scalar to accept the value would be easier (I believe this is how NumPy behaves)
- Empty scalar passed as value (e.g., `my_vector.apply(op.plus, empty_scalar)`)
  - C scalar raises before calling C function
  - `GrB_Scalar` works (but C function may raise if empty is invalid)

And, of course, `GrB_Scalar`s can be non-blocking.  This means that a GraphBLAS implementation is free to include scalars in expression trees that can later be optimized.

We can easily cast GrB `Scalars` to `Vectors` when using SuiteSparse:GraphBLAS.  This means we could implement e.g. `ewise_add`, `ewise_mult`, `ewise_union`, and `apply` on Scalars.  I don't know why these aren't in the spec to be honest.  Heh, I guess we could also allow Scalars in `gb.ss.concat` as length 1 Vectors or 1x1 Matrices depending on the output type.  We already do something similar to allow Vectors as `Nx1` matrices.

A potential "gotcha" for developers is there are now `._is_scalar` and `._is_cscalar` attributes.  I mistyped this once (that I know of) during this PR.

This PR also fixes/cleans up a couple minor edge cases.  I wouldn't be surprised if new ones got added ;)

PR #166, opened 2022-02-16
Autocompute `AmbiguousAssignOrExtract` as extract

### LINKED ISSUE #149: ENH: Auto-compute `AmbiguousAssignOrExtract` as extract
Just as we auto-compute expressions such as `A + B` in `(A + B) + C`, we can/should also auto-compute `AmbiguousAssignOrExtract` resulting from indexing such as `v[idx] + w`.

This requires that the object resulting from `A[idx]` know whether the return value is a scalar, vector, or matrix, and to have methods that match the output that will auto-compute.  In other words, we need to expand `AmbiguousAssignOrExtract` into multiple classes such as `ScalarIndexExpr`, `VectorIndexExpr`, `MatrixIndexExpr`, and to populate each class with methods from `grblas/_automethods.py`.

We should also improve the repr of each of these new classes.

Notably, `Assigner` and `Updater` don't need updated for auto-compute.

eriknw:
Fixes #149

This adds new classes such as `MatrixIndexExpr`, which are able to be auto-computed.  PR #157 made this relatively easy.  I still need to add a few tests.

### DIFF
--- a/grblas/_agg.py
+++ b/grblas/_agg.py
@@ -530,11 +530,11 @@ def _first_last(agg, updater, expr, *, in_composite, semiring):
         # Populate numpy array
         vals = np.empty(Is.size, dtype=A.dtype.np_type)
         for index, (i, j) in enumerate(zip(Is, Js)):
-            vals[index] = A[i, j].value
+            vals[index] = A[i, j].new().value
         # or Vector
         # v = expr._new_vector(A.dtype, size=A._nrows)
         # for i, j in zip(Is, Js):
-        #     v[i] = A[i, j].value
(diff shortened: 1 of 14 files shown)


eriknw:
@jim22k the Louvain notebook is failing _sometimes_.  I don't recall it ever being flaky before.  Would you be able to take a look at it when you have the chance?  Thanks.

In other news, it's nice to see the tests passing again when installing `psg` from source or wheels!

eriknw:
Alright, this is ready to go in!  Yet another fairly large PR that has backwards-incompatible changes.

PR #173, opened 2022-03-14
Update metadata to support SuiteSparse:GraphBLAS 6.2

eriknw:


eriknw:
Failures are expected (`python-suitesparse-graphblas` automated wheel building is broken).  Merging and releasing.

Issue #184, opened 2022-03-19
`Scalar.from_value()` segfaults when `value` type is unexpected


ParticularMiner:
I hit a segfault while running `test_apply_binary()` in `test_matrix.py`.

The bug may be reproduced as follows.

```python
import grblas
from grblas import binary
from grblas import Matrix


data = [
    [3, 0, 3, 5, 6, 0, 6, 1, 6, 2, 4, 1],
    [0, 1, 2, 2, 2, 3, 3, 4, 4, 5, 5, 6],
    [3, 2, 3, 1, 5, 3, 7, 8, 3, 1, 7, 4],
]
A = Matrix.from_values(*data)
A.apply(binary.plus, left=A)
```
A `TypeError` exception was supposed to have been raised.  Instead, it segfaults. 

The bug appears to originate from the following call:
```python
from grblas import Scalar


Scalar.from_value(A)
```

FYI, older `grblas` versions did not segfault on these calls.

eriknw:
Well, that's worrying.  Thanks for the report.

I'm unable to reproduce this.  What OS are you using @ParticularMiner?

ParticularMiner:
Using Windows 10.

eriknw:
These lines should exercise this, right?
https://github.com/metagraph-dev/grblas/blob/7dd54fd2f38d80a6cffa8a171182e1de929be54c/grblas/tests/test_matrix.py#L1065-L1068

ParticularMiner:
Yeah.  But they segfault all the same, even after updating SuiteSparse:GraphBLAS to the latest version (6.2.5).

eriknw:
I'm able to run tests on Windows in CI without segfaulting: https://github.com/metagraph-dev/grblas/pull/185 (still working to make everything pass).

Note that I didn't add any tests and am relying on the tests I referred to above.

This did point out that we probably need to test on Windows again.  My bad for ever dropping it.

eriknw:
@ParticularMiner can you try to see if you still see this error in the latest `python-graphblas`?  We don't have conda-forge packages for it yet.  Also, you should update to the latest `python-suitesparse-graphblas`.

ParticularMiner:
@eriknw @jim22k 

I've rechecked and found that this issue has been fixed in `python-graphblas`.  Exceptions are raised as expected and no more segfaults are being raised.  Thanks and well done!!  🎉

ParticularMiner:
[...]

PR #185, opened 2022-03-19
Investigate #184 by running on more OSes.

eriknw:
Also, don't show chained/nested exceptions when unnecessary.

### DIFF
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -147,12 +147,10 @@ jobs:
           auto-update-conda: true
           python-version: "3.9"
           channels: conda-forge
-          activate-environment: coveralls
-      - name: Update env
-        run: |
-          conda install -c conda-forge coveralls
-      - name: Coveralls Finished
+          activate-environment: finish_coveralls
+      - name: Finish coveralls
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          conda install -c conda-forge coveralls
           coveralls --finish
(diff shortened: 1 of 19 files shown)


eriknw:
Alright, apparently it's actually important for us to test on different OSes.  This test fixes many tests that were failing on Windows (don't get me started about `np.ldexp` on Windows!).

So, to keep our CI config matrix from exploding, I added a `pytest --randomly` which randomly chooses between blocking and non-blocking, record or not, runslow or not, and mapnumpy or not.  This give us 16 variations, and we run `pytest --randomly` 28 times in CI (so, on average, this will test 13.5 of 16 configurations for each set of runs).  Be on the lookout for intermittent test failures.  I also explicitly list out the matrix config options to run in CI.

The other bit of hygiene that this PR does is it adds `raise ... from None` more often so we don't chain exceptions when there's no value in doing so.

Unfortunately, I was unable to reproduce the segfault from #184, which was the impetus for this PR.

ParticularMiner:
Just to confirm: this latest merge has not fixed the segfault in #184.

eriknw:
> Just to confirm: this latest merge has not fixed the segfault in https://github.com/metagraph-dev/grblas/issues/184.

Probably not.  I didn't do anything that is specifically meant to try to fix it.  When you run tests locally, do you segfault?

ParticularMiner:
Interestingly, I do not segfault when I run tests locally.  But sometimes, when using the debugger, a segfault does prop up.  Whether a segfault occurs or not under the debugger appears to be arbitrary.  This is quite weird and I'll be very interested to know what could cause such behavior.

PR #186, opened 2022-03-20
added X.diag(k, ...) where type(X) in {Vector, Matrix}

### LINKED ISSUE #158: ENH: add GrB diag
We have SuiteSparse-specific `diag` methods such as `gb.ss.diag` and `my_matrix.ss.diag`.  These use GxB diag functions that take a scalar offset (for off-diagonals) and descriptors.

GrB diag is now part of the official spec.  We could add it as `gb.diag` and `my_matrix.diag`.

ParticularMiner:
Fixes #158 

Feel free to edit code in this PR if anything is amiss.


eriknw:
Cool, thanks for working on this @ParticularMiner!

Could you edit your initial comment to say e.g. "Fixes #158"?  With such a comment, GitHub is smart enough that it will close that issue when we merge this PR (it's also good practice to link PRs to existing issues if appropriate).

Also, I would expect this to call `GrB_Matrix_diag`.  This shouldn't call functions in `.ss` namespaces.

Apparently, the GraphBLAS spec only added the ability to assign a vector to the diagonal of a matrix.  I don't see any function to go the other way around.  Weird.

We currently have the following (which also have `k=0` optional argument):
- `my_vector.ss.diag(matrix)`: extract the diagonal of a matrix and assign it to `my_vector` discarding existing entries.
- `my_matrix.ss.diag(vector)`: assign values of `vector` to the diagonal of `my_matrix` and discard existing entries.
- `grblas.ss.diag(x)`: create a new vector or matrix depending on input type; "just works".

So, in typical GraphBLAS style, it's kind of weird that the API needs an already existing output object.  The third option above, `grblas.ss.diag(x)` is the most natural for most use cases.  So, for "GrB diag", I propose the following:
- `my_matrix.build_diag(vector)`: like the existing `my_matrix.ss.diag(vector)`
- `my_vector.diag()`: creates a _new_ Matrix with the diagonal set with `my_vector`.
  -  This should probably accept optional `name=`, `dtype=`, `nrows=`, and `ncols=` arguments (where `min(nrows, ncols) == my_vector.size`)

If we do this, then it probably makes sense to rename `my_matrix.ss.diag` to `my_matrix.ss.build_diag` to match, and add `my_matrix.ss.diag` that creates a new Vector.  We should do the same for `my_vector.ss` as well.

eriknw:
Oh man, I should look at docs before I write.  So, looking at the C API spec here https://graphblas.org/docs/GraphBLAS_API_C_v2.0.0.pdf it appears there are two main differences between `GrB_Matrix_diag` and `GxB_Matrix_diag`:
- `GrB_Matrix_diag` _creates_ a new, square matrix (if a non-square result is desired, I guess do `resize` afterwards)
- `GxB_Matrix_diag` accepts a descriptor

Surprisingly, SuiteSparse:GraphBLAS doesn't currently do the first.  The first argument is supposed to be a pointer to a `GrB_Matrix`, but SuiteSparse:GraphBLAS doesn't pass a pointer.  Hmph.  I guess this changes the API we should target.

Also, CC @jim22k for API review of GrB functions.  So, I guess we have a single function that accepts a Vector and creates a square Matrix.  Three options:
1. `A = my_vector.diag(k)`
2. `A = Matrix.diag(v, k)`
3. `A = grblas.diag(v, k)`

I vote for 1.

ParticularMiner:
Thank you @eriknw for your attention to detail!  I've removed the spec: `A.diag(...)` for Matrix `A` and put an informative error in its place.  

We can sort out the issues with `GrB_Matrix_diag()` vs. `GxB_Matrix_diag()` later when we get more information.

jim22k:
> Three options:
> 
> 1. `A = my_vector.diag(k)`
> 2. `A = Matrix.diag(v, k)`
> 3. `A = grblas.diag(v, k)`

I like 1 and 3. 2 is definitely out. I would instead change 2 to be `A = Matrix.from_diag(v, k)`. That would be acceptable to me.

As far as the spec only covering Vector->Matrix diagonals, I think we should also handle Matrix->Vector. `grblas` should always be a superset of the existing functionality declared in the spec, as long as there are extension methods or recipes that we can target. We already do this with `extract_row`, which is just `extract_column` with a transpose bit set. For the Matrix->Vector recipe, we could select with `GrB_DIAG`, then reduce to vector with any monoid.

jim22k (approved):

eriknw:
Thanks for the feedback @jim22k !  Yeah, good point about being able to make a GrB-only recipe for taking the diagonal of a Matrix (once we have select).  This is probably why it's not in the spec as a separate method.

So, +1 for:
- `my_matrix.diag(k) -> Vector`
- `my_vector.diag(k) -> Matrix`

These should both call `self.ss.diag` for now (probably).

Regarding `grblas.diag`, we don't yet have any top-level functions like this.  I'm hesitant to put it in the top-level namespace, because there's likely going to be pressure to add even more functions similar to this, and I don't like things getting too cluttered.  Like `gb.ss.diag`, I think we could have other namespace(s) for e.g. `gb.xxx.diag`.  I don't know what we would call it.  `gb.grb.diag`? `gb.utils.diag`? `gb.funcs.diag`? `gb.core.diag`?  We already put `gb.io` functions in their own namespace.  On the other hand, maybe there won't be that many top-level functions such as this.  `gb.ss` only has two: `diag` and `concat`.  `concat` is pretty handy--we could make grb-only recipes for it too.

jim22k:
All this code is already part of PR #188, so closing.

PR #188, opened 2022-03-23
improved portability of first/last Aggregators and restored Matrix.diag()

ParticularMiner:
- This PR fixes "first" and "last" aggregators so that they can also support `dask_grblas`.  
- It also restores the specification: `my_matrix.diag(k, ...)` (extraction of a diagonal from a `Matrix` to a `Vector`) which helps enable the support for `dask-grblas`.   This specification was removed earlier in PR #186 .

Feel free to edit this PR if anything is amiss!


jim22k (approved):

eriknw:
Hooray, thanks @ParticularMiner 🎉

grblas/_agg.py
```diff
@@ -517,35 +511,32 @@ def _argminmax(agg, updater, expr, *, in_composite, monoid):
...
     elif expr.cfunc_name.startswith("GrB_Vector_reduce"):
         v = expr.args[0]
         init = expr._new_matrix(bool, nrows=v._size, ncols=1)
         init[...] = False  # O(1) dense matrix in SuiteSparse 5
-        step1 = semiring(v @ init).new()
+        step1 = semiring_(v @ init).new()
         index = step1[0].new().value
-        if index is None:
+        # `==` instead of `is` automatically triggers index.compute() in dask-grblas:
+        if index == None:  # noqa
```
eriknw, line 539: We might want to consider supporting e.g. `t = v[s].new()`  where `s` and `t` are both `gb.Scalar`s that are empty.  Currently, scalar indices are `GrB_Index` type, not `GrB_Scalar` type, so this isn't supported by SuiteSparse:GraphBLAS.  But, we could support this in `grblas` and `dask-grblas`, and doing so would allow us to avoid `dask-grblas` computing during `index == None`.

ParticularMiner: Great idea from an optimization standpoint!   

Heh, this automatically means also dealing with lists of indices of different types, like `v[2 s 1 None t].new()`.

PR #190, opened 2022-03-30
Deprecation cycle!

eriknw:
- Remove e.g. `reduce_rows` (was already deprecated)
- Deprecate e.g. `v.ss.diag(A)` and rename to `v.ss.build_diag(A)` instead
- Remove `grblas.backends`

As was discussed earlier today.


eriknw:
This still has `gb.ss.diag`, but deprecating this would have been more work.  Maybe later.  Failing tests are coveralls timeouts.  Merging.

PR #191, opened 2022-03-31
Add `rpow` and improve tab-completion of operator namespaces.

eriknw:
Also, make sure all commute targets exists (but haven't done this for numpy namespaces yet)


eriknw:
Wow, this seems to have significantly impacted import time.  Hmph.  Let me try to make things a bit more lazy.

eriknw:
Whew, looks like import time is faster again.  I'm looking at the time for "Verify build" section of the Github Actions, which makes sure we can import everything.

PR #192, opened 2022-03-31
Autocompute scalar expressions in `.update` if valid.

eriknw:
This was an interesting edge case that I stumbled upon!
Anyway, I think this can also provide better error messages.

### DIFF
--- a/grblas/base.py
+++ b/grblas/base.py
@@ -420,6 +420,16 @@ class BaseType:
                         updater[...] = scalar
                     return
 
+        if type(self) is not expr.output_type:
+            if expr.output_type._is_scalar and config.get("autocompute"):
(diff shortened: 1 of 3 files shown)


grblas/base.py
```diff
@@ -420,6 +420,16 @@ def _update(self, expr, mask=None, accum=None, replace=False, input_mask=None):
                         updater[...] = scalar
                     return
 
+        if type(self) is not expr.output_type:
+            if expr.output_type._is_scalar and config.get("autocompute"):
+                self._update(expr._get_value(), mask, accum, replace, input_mask)
+                return
+            from .scalar import Scalar
+
+            valid = (Scalar, type(self), type(type(self).__name__ + "Expression", (), {}))
```
eriknw, line 429: Just don't ask any questions about the third element in `valid`! 😛

PR #194, opened 2022-04-03
Add `x.ss.iso_value` to Vector and Matrix objects

eriknw:
I think it's nice to capture intent here even though this is currently a recipe.  We could make this faster by e.g. using iterators that SuiteSparse:GraphBLAS recently added.

Some possible alternative behaviors of `x.ss.iso_value`:
- If empty, return empty scalar (it currently raises)
- If not iso-valued, return `None` (it currently raises)

### DIFF
--- a/grblas/_ss/matrix.py
+++ b/grblas/_ss/matrix.py
@@ -352,6 +352,12 @@ class ss:
         return is_iso[0]
 
     @property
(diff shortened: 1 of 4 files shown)


eriknw:
Merging so I can do a release.  Feel free to share thoughts if you think this should behave differently.

PR #195, opened 2022-04-03
Improve `gb.io.from_networkx`

eriknw:


eriknw:
The documentation could probably stand to be improved (*cough* *cough*), and I'm curious about how to convert to UDTs, but I want to get a release out the door, so merging.

PR #201, opened 2022-04-05
Add `x.ss.iter{keys,values,items}`

eriknw:
I've been sitting on this for a while, so I thought I'd push this up.
I still need to read the documentation more closely.  Also,
SuiteSparse:GraphBLAS provides even more iteration functionality.

This should make `x.ss.iso_value` even faster, which can speed up PageRank :tada: . Speaking of, this changes the return type of `x.ss.iso_value` from `Scalar` to a Python scalar.  Meh.


eriknw:
Alright, I think this is finished and ready for review.

This doesn't expose row iterators and column iterators for matrices, but I'm not really sure what those would be good for.  So, I think it's fine to punt for now.

jim22k (approved):

PR #202, opened 2022-04-06
Broadcast vectors across rows with `plus(A, v)`.

### LINKED ISSUE #197: Add more convenient way to use diag to broadcast Vector across Matrix rows or columns
Diag is super-handy.  We use it regularly.  Can we make it easier to use to broadcast vectors across matrix rows or columns?  @jim22k, I believe you've expressed this desire before.

For example, we often do this:
```python
op.any_plus(A @ v.diag())
```
This is mostly fine.  It spells out exactly how to do it.  But, the `any_` monoid in the semiring adds no cognitive value (other than to say "hey, this is a semiring").  Also, one may wonder "does this broadcast values like I expect", since it's not how one would normally do this in NumPy (which would be `A + v`).

Here's an idea.  Since `plus(A, s)` already "broadcasts" the scalar `s` across the entries of the Matrix, how about `plus(A, v)` as shorthand for `any_plus(A @ v.diag())`?  I actually think this is quite nice.  This works whether the Vector is on the left or right, and given a Matrix or TransposedMatrix.  `plus(v, v)` should still raise since nothing is being broadcast, right?

Perhaps we could also add a method to Matrix and/or Vectors, such as `A.mbv(v)`.  I don't have a good name for this.

eriknw:
Fixes #197.  This is the most narrow implementation as described in #197.

Allowed:
- `plus(A, v)` -> `any_plus(A @ v.diag())`
- `plus(v, A)` -> `any_plus(v.diag() @ A)`

Not allowed:
- `min_plus(A, v)`
- `A + v`
- `v + A`
- `A += v`

Need to update error messages before merging (pending approval).

### DIFF
--- a/grblas/_infixmethods.py
+++ b/grblas/_infixmethods.py
@@ -2,6 +2,7 @@ from . import binary, unary
 from .dtypes import BOOL
 from .infix import MatrixInfixExpr, VectorInfixExpr
 from .matrix import Matrix, MatrixExpression, MatrixIndexExpr, TransposedMatrix
+from .operator import _call_op
 from .utils import output_type
 from .vector import Vector, VectorExpression, VectorIndexExpr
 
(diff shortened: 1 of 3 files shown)


eriknw:
Closing in favor of https://github.com/metagraph-dev/grblas/pull/203

PR #207, opened 2022-04-11
Add `select` Operation

jim22k:
This implements item 1 in the [proposal for adding select](https://github.com/metagraph-dev/grblas/issues/147).

- Put the IndexUnaryOps which return bool into the grblas.select namespace
- Allow these to be called directly (gb.select.tril(A)) to perform a select operation
- Include the "special" versions (select.row, select.column, select.value), but also include the "raw" versions (select.rowle, etc)
- Add a select method (e.g. A.select("tril"), A.select("<", 1), A.select(gb.select.diag, 1), A.select("row<=", 1))


eriknw:
Aha, I see you found errors in the bizarro tests: https://github.com/metagraph-dev/grblas/blob/8e9acb9c1ac0e5689fc2f1066fd60020ea1c3ffe/.github/workflows/test_and_build.yml#L102-L109

This trick is kinda weird, but it's the best way I know to test both types of scalars (C and GrB) throughout the API.  You may find it handy to switch to bizarro world locally with this command:
```bash
#!/bin/bash

find grblas -type f -name "*.py" -print0 | xargs -0 sed -i -s \
    -e '/# pragma: is_grbscalar/! s/is_cscalar=False/is_cscalar=True/g' \
    -e '/# pragma: is_grbscalar/! s/is_cscalar = False/is_cscalar = True/g' \
    -e '/# pragma: to_grb/ s/is_cscalar=True/is_cscalar=False/g' \
    -e '/# pragma: to_grb/ s/is_cscalar = True/is_cscalar = False/g'
```
Scalars are GrB by default, but there are a few places in the code where they must be a specific type of scalar.
- When creating a scalar with `is_cscalar=True`, it is assumed that it must be a C scalar.
  - Add `# pragma: to_grb` if it is allowed to be GrB
- When creating a scalar with `is_cscalar=False`, it is assumed it is also safe to change it to a C scalar.
  - Add `# pragma: is_grbscalar` if it *must* be a GrB scalar

grblas/matrix.py
```diff
@@ -785,6 +785,39 @@ def apply(self, op, right=None, *, left=None):
...
+            except TypeError:
+                thunk = self._expect_type(
+                    thunk,
+                    Scalar,
+                    within=method_name,
+                    keyword_name="thunk",
+                    extra_message="Literal scalars also accepted.",
+                    op=op,
+                )
+        cfunc_name = f"GrB_Matrix_select_{self.dtype}"
```
eriknw, line 809: The dtype in the function is based on the dtype of the thunk value, right?  Also, this needs updated to handle UDTs.

grblas/operator.py
```diff
@@ -785,7 +800,7 @@ def _initialize(cls):
                                 obj = cls(name, is_positional=name in cls._positional)
                             setattr(cls._module, name, obj)
                             _STANDARD_OPERATOR_NAMES.add(f"{cls._modname}.{name}")
-                            if not hasattr(op, name):
+                            if merge_namespace and not hasattr(op, name):
```
eriknw, line 803: What does this actually change?

grblas/operator.py
```diff
@@ -186,6 +189,18 @@ def __call__(self, val):
         )
 
 
+class TypedBuiltinIndexUnaryOp(TypedOpBase):
+    __slots__ = ()
+    opclass = "IndexUnaryOp"
+
+    def __call__(self, val, thunk=None):
```
eriknw, line 196: Can we set the default to `thunk=0` here?

grblas/operator.py
```diff
@@ -186,6 +189,18 @@ def __call__(self, val):
...
+        if isinstance(val, InfixExprBase):
```
eriknw, line 197: Can we drop this check?  For example, `gb.select.tril(A @ B)` should be fine, right?  We use `output_type(val)` instead of `type(val)` to handle expressions and objects alike.

grblas/matrix.py
eriknw: I'm not super-clear yet on the dtype handling, but the function with the dtype in the name is only for C scalars.  We should use `GrB_Matrix_select_Scalar` if the thunk is a GrB scalar.  I bet this would help the failing tests too.

grblas/select/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+            f"Typical usage: select.{callname}(x <= 5)"
+        )
+    tensor = expr.args[0]
+    thunk = expr.args[1]
+    method = f"{opname}{expr.op.name}"
+    if method not in globals():
+        # Attempt to convert < into <= (rowlt is not part of official spec, but rowle is)
+        if expr.op.name == "lt":
+            method = f"{opname}le"
+            thunk.value -= 1
```
eriknw, line 37: Any risk of `thunk` being an unsigned integer with value 0 here (and below)?

grblas/operator.py
```diff
@@ -1095,6 +1110,52 @@ def __reduce__(self):
...
+        ],
+    }
+
+    @classmethod
+    def _initialize(cls):
+        super()._initialize(merge_namespace=False)
+        # Update type information to handle more than bool
+        # fmt: off
+        bool_names = ("tril", "triu", "diag", "offdiag", "colle", "colgt", "rowle", "rowgt")
+        types = (INT8, UINT8, INT16, UINT16, INT32, UINT32, INT64, UINT64, FP32, FP64, FC32, FC64)
```
eriknw, line 1143: This may cause a problem on Windows if `FC32` and `FC64` don't exist.

eriknw:
We added `my_op.is_positional` attribute to operators for ones like `positioni` and `firsti`.  I think IndexUnaryOps should also have `.is_positional` attribute, and I think it should be true for all IndexUnaryOps except those with `"VALUE"` in the name.  Then, in `select`, the thunk should be treated as int64 if the op is positional unless the int32 version of the operator was explicitly given.

jim22k (commented):
I'm going to change IndexUnaryOp to SelectOp. When we add apply with index unary, I will add that in as a different op class.

graphblas/vector.py
```diff
@@ -708,6 +708,49 @@ def apply(self, op, right=None, *, left=None):
...
+        """
+        GrB_Vector_select
+        Compute SelectOp at each element of the calling Vector, keeping
+        elements which return True.
+        """
+        method_name = "select"
+        if thunk is None:
+            thunk = False  # most basic form of 0 when unifying dtypes
+        if type(thunk) is not Scalar:
+            dtype = self.dtype if self.dtype._is_udt else None
```
eriknw, line 721: We should only use this pattern if the op is not positional.  A UDT is not the type we want when comparing against indices!

graphblas/vector.py
```diff
@@ -708,6 +708,49 @@ def apply(self, op, right=None, *, left=None):
...
+            except TypeError:
+                thunk = self._expect_type(
+                    thunk,
+                    Scalar,
+                    within=method_name,
+                    keyword_name="thunk",
+                    extra_message="Literal scalars also accepted.",
+                    op=op,
+                )
+        op = get_typed_op(op, self.dtype, thunk.dtype, is_right_scalar=True, kind="select")
```
eriknw, line 733: We should only include `self.dtype` here when the operator is not positional (similar as prev comment).

eriknw:
Please try to increase coverage: https://coveralls.io/builds/48230528

jim22k:
@eriknw I think this is ready. Please take one last look, and if you're happy with it, go ahead and merge.

graphblas/matrix.py
```diff
@@ -791,6 +791,52 @@ def apply(self, op, right=None, *, left=None):
...
+        return MatrixExpression(
+            method_name,
+            cfunc_name,
+            [self, thunk],
+            op=op,
+            expr_repr="{0.name}.select({op}, thunk={1._expr_name})",
+            nrows=self._nrows,
+            ncols=self._ncols,
+            dtype=self.dtype,
+        )
```
eriknw, line 838: This doesn't yet work on transposed matrices.  Need to add `at=self._is_transposed` to the MatrixExpression.  Also, should probably add a test to exercise this.

graphblas/operator.py
```diff
@@ -1092,6 +1109,57 @@ def __reduce__(self):
     __call__ = TypedBuiltinUnaryOp.__call__
 
 
+class SelectOp(OpBase):
```
eriknw, line 1112: This class needs `__slots__`.  Probably `__slots__ = "is_positional"`.

graphblas/operator.py
```diff
@@ -2475,6 +2546,18 @@ def get_semiring(monoid, binaryop, name=None):
...
+    ">": select.valuegt,
+    "<=": select.valuele,
+    ">=": select.valuege,
+    "!=": select.valuene,
+    "==": select.valueeq,
+    "col<=": select.colle,
+    "col>": select.colgt,
+    "row<=": select.rowle,
+    "row>": select.rowgt,
+}
```
eriknw, line 2560: Can we also add `"col<"`, `"row<"`, `"col>="`, and `"row>="` with recipes to make them work without too much hassle?

graphblas/select/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        rv = func(**kwargs)
+        globals()[key] = rv
+        return rv
+    raise AttributeError(f"module {__name__!r} has no attribute {key!r}")
+
+
+def _resolve_expr(expr, callname, opname):
+    from graphblas.base import BaseExpression
+
+    if not isinstance(expr, BaseExpression) and hasattr(expr, "select"):
```
eriknw, line 28: What's with `and hasattr(expr, "select")` here?  Did you mean `or not hasattr(expr, "select")`?  It may be clearer to check explicitly for `MatrixExpression` and `VectorExpression`.

graphblas/select/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        elif expr.op.name == "ge":
+            method = f"{opname}gt"
+            thunk -= 1
+        if method not in globals():
+            raise ValueError(f"Unknown or unregistered select method: {method}")
+    return globals()[method](tensor, thunk)
+
+
+def value(expr):
+    return _resolve_expr(expr, "value", "value")
```
eriknw, line 53: It would be super-nice to have docstrings for these three functions--even super-basic ones that showed example usage.

graphblas/select/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        raise TypeError(
+            f"Expected VectorExpression or MatrixExpression; found {type(expr)}\n"
+            f"Typical usage: select.{callname}(x <= 5)"
+        )
+    tensor = expr.args[0]
+    thunk = expr.args[1]
+    method = f"{opname}{expr.op.name}"
+    if method not in globals():
+        # Convert thunk to Python int to avoid possible subtraction with uints
+        thunk = int(thunk.value)
```
eriknw, line 38: No need to cast to `int` here.  `thunk.value` is a Python scalar.

eriknw (commented):
It occurs to me that several of the SelectOps are awkward for Vectors.  `gb.select.row` needs to be used instead of `gb.select.column`, but column-wise selectors are technically valid on Vectors, but they don't do much.   Similarly, `tril` is kind of useless.  `diag` and `offdiag` can be use to select a specific index or all except a specific index, but the thunk to indicate which index needs to be negative.

Should we consider having type-specific (Vector, Matrix) SelectOps?  There's definitely some awkwardness around UnaryOps/SelectOps, isn't there?

This is super-close to merging.  I'll nag you about coverage on the next select PR ;)

eriknw:
Also, I tried select on a Matrix with a UDT, and it worked!  Can you add a simple example to `test_udt` in "test_matrix.py"?  For example, you could verify that these give the same result: `gb.unary.one(gb.select.tril(M))` and `gb.select.tril(gb.unary.one(M))`.

eriknw:
For warm-fuzzies, you may want to merge with main branch and update to `python-suitesparse-graphblas` 7.0.3.1.

graphblas/select/__init__.py
jim22k: I was trying to catch `ScalarExpression`, but you're right that explicitly calling out `MatrixExpression` and `VectorExpression` will be more readable.

graphblas/matrix.py
jim22k: Good catch

graphblas/operator.py
jim22k: Let's hold off on this for now. There is no easy way to make this happen. Hopefully with user-defined IndexUnaryOps, this will have an easier solution.

eriknw:
This is in, thanks so much @jim22k :tada:

PR #210, opened 2022-04-14
Complex monoid UDFs are finally working!

eriknw:
Update to latest `python-suitesparse-graphblas` to avoid segfaults 🤞 .
Also, let's see if using `pytest-randomly` works now.

### DIFF
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -54,7 +54,7 @@ jobs:
           activate-environment: graphblas
       - name: Update env
         run: |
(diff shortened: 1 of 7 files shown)


eriknw:
I ran two sets of CI tests with `pytest-randomly`, and everything passed!  I think we may have more-or-less fixed the segfaults.  The complex monoid segfault has been known for a couple of years, and it feels really good to finally get it fixed.  The solution still seems super-weird to me 🤷‍♂️ 

I deleted `pytest-randomly` for now.  May add later.  Seems like a pretty good trick for testing robustness.

Everybody should update to `python-suitesparse-graphblas` 7.0.3.1!

PR #215, opened 2022-04-16
Add `x.reposition` to "shift" values around.

### LINKED ISSUE #180: Add pad/shift/translate
There is already `.resize`, which can add or trim an object on the right or the bottom.  It might be handy to have a more general operation that can both resize objects and shift values (or only shift).

For example, such a function could be used to translate a small local chunk in `dask-grblas` to its proper place in the global matrix.  This might be handy to implement https://github.com/metagraph-dev/dask-grblas/issues/19.

We should also consider how this this would look as a C API and Python API.  Some questions:
- Should this mutate an existing object like `.resize` or create a new object?
- What should the Python API be?
  - Should we try to mimic `numpy.pad`?
    - `pad(2)` to add a length 2 border around everything
    - `pad((1, 2))` to pad with width 1 on the left and width 2 on the right
    - Should we allow negative such as `pad(-2)` to undo `pad(2)` or `pad((1, -1))` to shift everything to the right by 1 but keep the shape the same?
- Possible names: `pad`, `shift`, `translate`, etc.
- If we "shift" all values by N, is there any value in having different boundary conditions such as drop or wraparound?

I think there are several possible variations here, and I don't yet have a strong opinion.  So, let's discuss!

This function may come in handy:
https://github.com/numpy/numpy/blob/a8f9711493adee93fa3d61e7ef1bee11d7055a85/numpy/lib/arraypad.py#L454

Also, thanks to @ParticularMiner for first asking about this in a side channel.

eriknw:
Fixes #180

@jim22k and @ParticularMiner I think this is what we discussed.  Please take a look.

This is our first "unofficial" method on `Matrix` objects directly (`Vector` already had `inner` and `outer`).  I think it's justified: although simple, it would be easy to get wrong or do poorly.  It's also useful enough, and we have a good name and arguments for it.


eriknw:
I modified `x.reposition` to return an expression.  I think this is better, because it lets us use the same patterns as elsewhere and efficiently applies masks, accumulators, etc.

In general, I think custom recipes that we add should return expressions and not objects if there's a reasonable way to do so.  Recipes that rely heavily on Numba such as `M.ss.compactify_rowwise` don't need to return expressions.

Hmm, I think I might be able to use the conventions added in this PR to make broadcasting vectors across matrices even better for `ewise_add` and `ewise_union` when there are masks.  Also, better reprs.

jim22k (approved):

PR #216, opened 2022-04-18
Auto-register dtypes as UDTs.

eriknw:
Also, only have a single object for each `numpy.dtype` for UDTs.  This makes it easier and less awkward for us, but could result in weirdness if the same dtype is registered multiple times under different names.  Probably okay.


eriknw:
One small step towards better and more convenient UDTs.  Merging.

PR #218, opened 2022-04-18
Better default names for UDTs.

eriknw:
This doesn't have as much detail as numpy or numba typically show,
but we want the default name to be friendlier since it shows up in object reprs.
We still might want to update object reprs to better handle long UDT names.

### DIFF
--- a/graphblas/dtypes.py
+++ b/graphblas/dtypes.py
@@ -99,7 +99,7 @@ def register_anonymous(dtype, name=None):
         check_status_carg(status, "Type", gb_obj[0])
     # For now, let's use "opaque" unsigned bytes for the c type.
     if name is None:
-        name = repr(dtype)
+        name = _default_name(dtype)
     numba_type = numba.typeof(dtype).dtype
     rv = DataType(name, gb_obj, None, f"uint8_t[{dtype.itemsize}]", numba_type, dtype)
     _registry[gb_obj] = rv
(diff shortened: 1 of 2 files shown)


graphblas/dtypes.py
```diff
@@ -262,3 +262,21 @@ def unify(type1, type2, *, is_left_scalar=False, is_right_scalar=False):
...
+        subdtype = _default_name(dtype.subdtype[0])
+        shape = ", ".join(map(str, dtype.subdtype[1]))
+        return f"{subdtype}[{shape}]"
+    elif dtype.names:
+        args = ", ".join(
+            f"{name!r}: {_default_name(dtype.fields[name][0])}" for name in dtype.names
+        )
+        return "{%s}" % args
+    else:
+        return repr(dtype)
```
eriknw, line 282: We could also return `dtype.str` here, which would show `|S5` instead of `dtype('S5')`.

graphblas/tests/test_dtype.py
```diff
@@ -194,3 +194,15 @@ def test_auto_register():
...
+def test_default_names():
+    from graphblas.dtypes import _default_name
+
+    assert _default_name(np.dtype([("x", np.int32), ("y", np.float64)], align=True)) == (
+        "{'x': INT32, 'y': FP64}"
+    )
+    assert _default_name(np.dtype("(29,)uint8")) == "UINT8[29]"
+    assert _default_name(np.dtype("(3,4)bool")) == "BOOL[3, 4]"
+    assert _default_name(np.dtype((np.dtype("(5,)float64"), (6,)))) == "FP64[5][6]"
+    assert _default_name(np.dtype("S5")) == "dtype('S5')"
```
eriknw, line 208: These seem reasonable to me, but I'm open to suggestions.  For example, `Record(x=INT32, y=FP64)`.

PR #223, opened 2022-04-23
Add `gb.binary.binom` to compute binomial coefficients.

eriknw:
This is needed to calculate transitivity of a graph.

This is a "nice" example that shows how to customize creation of UDFs.  On the other hand, it shows how bespoke and sensitive it is to do.

I wonder whether the return type should be `INT64` or `FP64`, since binomial coefficients can get quite large in general.  Or maybe we should add `FP64` to the valid types (input and output).

### DIFF
--- a/graphblas/operator.py
+++ b/graphblas/operator.py
@@ -1190,6 +1190,43 @@ def _isclose(rel_tol=1e-7, abs_tol=0.0):
     return inner
 
 
+_MAX_INT64 = np.iinfo(np.int64).max
+
+
+def _binom(N, k):  # pragma: no cover
+    # Returns 0 if overflow or otherwise bad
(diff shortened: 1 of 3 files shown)


eriknw:
Adding approximate binomial coefficients gets a lot more complicated, so I'd like to punt on that in this PR.  But, if we ever do want to support approximate binomial coefficients, I would want this as `binom[FP64]`.

eriknw:
We don't actually _need_ this for transitivity ([see here](https://github.com/python-graphblas/graphblas-algorithms/pull/3/files#diff-ac7b8544319ec5529f77c7691931c892674075c4373dbba514285e149a03e915R127)).

I'm still okay with adding this.  It serves as a (messy) example for how to customize creation of a function, including loading lazily.

PR #224, opened 2022-04-25
Add `sort=True` keyword to `to_values`.

eriknw:
I find this especially true for Vectors.  SuiteSparse:GraphBLAS often sorts things, but this isn't guaranteed.  So, let's make it explicit, and promise to sort manually for future backends if necessary.

### DIFF
--- a/graphblas/matrix.py
+++ b/graphblas/matrix.py
@@ -1736,8 +1736,8 @@ class TransposedMatrix:
         return self._matrix.dtype
 
     @wrapdoc(Matrix.to_values)
-    def to_values(self, dtype=None):
-        rows, cols, vals = self._matrix.to_values(dtype)
+    def to_values(self, dtype=None, *, sort=True):
+        rows, cols, vals = self._matrix.to_values(dtype, sort=sort)
         return cols, rows, vals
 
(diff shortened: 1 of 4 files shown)


eriknw:
@jim22k any violent reactions to this?  In the easy case where the `sort=` argument isn't provided, this can result in extra unnecessary computation (probably fine).  There's also the potential confusion when exporting matrices in rowwise or columnwise format, which determines how they're sorted in SuiteSparse:GraphBLAS.

In argument _for_ this PR, the results are often sorted so that it's easy to assume they will always be and create tests that pass, and I've written such algorithms and tests.  If you're tuning an algorithm and know you don't need sorted indices, it's nice and explicit to pass `sort=False`.

jim22k:
My only reaction is that `sort=True` is the default, but it raises an error for anything non-SuiteSparse. While that is fine today, it isn't a good long term approach.

But I do like the guarantee of sorted indices. That feels right.

jim22k (approved):

eriknw:
> My only reaction is that sort=True is the default, but it raises an error for anything non-SuiteSparse. While that is fine today, it isn't a good long term approach.

Yeah, it does sign us up for future work.

PR #227, opened 2022-05-04
Allow record dtypes to be created more easily via dicts

eriknw:
For example, with this, one can use `{'x': int, 'y': float}` as a dtype, which is a lot simpler than `np.dtype([("x", np.int64), ("y", np.float64)])`.  I wonder whether there are any "gotchas" with this.


eriknw:
I think this is nice, so merging.  YOLO.

PR #229, opened 2022-05-05
Make apply accept IndexUnaryOps

jim22k:
- Add IndexUnaryOp and typed versions
- New gb.indexunary namespace
- Apply works with both IndexUnaryOp and SelectOp
  - The only real difference is what happens when you call those ops. Passing them to `apply` is treated identically when calling into SuiteSparse.
- Add tests

### DIFF
--- a/graphblas/matrix.py
+++ b/graphblas/matrix.py
@@ -10,7 +10,7 @@ from .dtypes import _INDEX, FP64, lookup_dtype, unify
 from .exceptions import DimensionMismatch, NoValue, check_status
 from .expr import AmbiguousAssignOrExtract, IndexerResolver, Updater
 from .mask import StructuralMask, ValueMask
-from .operator import get_semiring, get_typed_op
+from .operator import find_opclass, get_semiring, get_typed_op, op_from_string
 from .scalar import _MATERIALIZE, Scalar, ScalarExpression, _as_scalar, _scalar_index
 from .utils import (
     _CArray,
@@ -740,11 +740,23 @@ class Matrix(BaseType):
         Apply UnaryOp to each element of the calling Matrix
         A BinaryOp can also be applied if a scalar is passed in as `left` or `right`,
             effectively converting a BinaryOp into a UnaryOp
+        An IndexUnaryOp can also be applied with the thunk passed in as `right`
         """
         method_name = "apply"
         extra_message = (
-            "apply only accepts UnaryOp with no scalars or BinaryOp with `left` or `right` scalar."
+            "apply only accepts UnaryOp with no scalars or BinaryOp with `left` or `right` scalar "
+            "or IndexUnaryOp with `right` thunk."
         )
+        if isinstance(op, str):
+            op = op_from_string(op)
+        opclass = find_opclass(op)[1]
+        if opclass in ("IndexUnaryOp", "SelectOp"):
+            # Provide default value for index unary
+            if right is None:
+                right = False  # most basic form of 0 when unifying dtypes
+            if left is not None:
+                raise TypeError("Do not pass `left` when applying IndexUnaryOp")
+
         if left is None and right is None:
             op = get_typed_op(op, self.dtype, kind="unary")
             self._expect_op(
@@ -807,13 +819,20 @@ class Matrix(BaseType):
                         extra_message="Literal scalars also accepted.",
                         op=op,
                     )
-            op = get_typed_op(op, self.dtype, right.dtype, is_right_scalar=True, kind="binary")
+            if opclass in ("IndexUnaryOp", "SelectOp"):
+                op = get_typed_op(
+                    op, self.dtype, right.dtype, is_right_scalar=True, kind="indexunary"
+                )
+                cfunc_method = "IndexOp"
+            else:
+                op = get_typed_op(op, self.dtype, right.dtype, is_right_scalar=True, kind="binary")
+                cfunc_method = "BinaryOp2nd"
             if op.opclass == "Monoid":
                 op = op.binaryop
             else:
                 self._expect_op(
                     op,
-                    "BinaryOp",
+                    ("BinaryOp", "IndexUnaryOp", "SelectOp"),
                     within=method_name,
                     argname="op",
                     extra_message=extra_message,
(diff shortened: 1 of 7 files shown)


jim22k:
This is the minimum effort to make this work. A future PR will address user-defined versions for both IndexUnaryOp and SelectOp.

However, @eriknw if you feel strongly that we need to fix the way things work for Vectors (i.e. must use rowindex, rowle, etc) as part of this PR, I am okay adding that. Or we can push it to a future PR as well.

graphblas/matrix.py
```diff
@@ -740,11 +740,23 @@ def apply(self, op, right=None, *, left=None):
...
         """
         method_name = "apply"
         extra_message = (
-            "apply only accepts UnaryOp with no scalars or BinaryOp with `left` or `right` scalar."
+            "apply only accepts UnaryOp with no scalars or BinaryOp with `left` or `right` scalar "
+            "or IndexUnaryOp with `right` thunk."
         )
+        if isinstance(op, str):
+            op = op_from_string(op)
+        opclass = find_opclass(op)[1]
```
eriknw, line 752: Might as well set `op` here too (for, e.g., parameterized operators).

graphblas/matrix.py
```diff
@@ -740,11 +740,23 @@ def apply(self, op, right=None, *, left=None):
...
+        if opclass in ("IndexUnaryOp", "SelectOp"):
```
eriknw, line 753: Small nit: I now prefer sets instead of tuples, so like ` in {"IndexUnaryOp", "SelectOp"}`

graphblas/matrix.py
```diff
@@ -807,13 +819,20 @@ def apply(self, op, right=None, *, left=None):
...
-            op = get_typed_op(op, self.dtype, right.dtype, is_right_scalar=True, kind="binary")
+            if opclass in ("IndexUnaryOp", "SelectOp"):
+                op = get_typed_op(
+                    op, self.dtype, right.dtype, is_right_scalar=True, kind="indexunary"
+                )
+                cfunc_method = "IndexOp"
+            else:
+                op = get_typed_op(op, self.dtype, right.dtype, is_right_scalar=True, kind="binary")
+                cfunc_method = "BinaryOp2nd"
             if op.opclass == "Monoid":
```
eriknw, line 830: This if-else block can be indented (no need to check if opclass is an select op).  Also, we can just use `opclass` now instead of `op.opclass`.

graphblas/operator.py
```diff
@@ -1110,6 +1120,68 @@ def __reduce__(self):
     __call__ = TypedBuiltinUnaryOp.__call__
 
 
+class IndexUnaryOp(OpBase):
+    __slots__ = ("is_positional",)
```
eriknw, line 1124: Small nit: no need to have this as a tuple (same for `__slots__` of `SelectOp`).

graphblas/vector.py
```diff
@@ -661,8 +661,19 @@ def apply(self, op, right=None, *, left=None):
         """
         method_name = "apply"
         extra_message = (
-            "apply only accepts UnaryOp with no scalars or BinaryOp with `left` or `right` scalar."
+            "apply only accepts UnaryOp with no scalars or BinaryOp with `left` or `right` scalar"
+            "or IndexUnaryOp with `right` thunk."
         )
+        if isinstance(op, str):
+            op = op_from_string(op)
+        opclass = find_opclass(op)[1]
```
eriknw, line 669: and here

graphblas/vector.py
```diff
@@ -661,8 +661,19 @@ def apply(self, op, right=None, *, left=None):
...
+        if opclass in ("IndexUnaryOp", "SelectOp"):
```
eriknw, line 670: and here

graphblas/matrix.py
```diff
@@ -807,13 +819,20 @@ def apply(self, op, right=None, *, left=None):
                         extra_message="Literal scalars also accepted.",
                         op=op,
                     )
-            op = get_typed_op(op, self.dtype, right.dtype, is_right_scalar=True, kind="binary")
+            if opclass in ("IndexUnaryOp", "SelectOp"):
```
eriknw, line 822: and here

graphblas/vector.py
```diff
@@ -725,13 +736,20 @@ def apply(self, op, right=None, *, left=None):
                         extra_message="Literal scalars also accepted.",
                         op=op,
                     )
-            op = get_typed_op(op, self.dtype, right.dtype, is_right_scalar=True, kind="binary")
+            if opclass in ("IndexUnaryOp", "SelectOp"):
```
eriknw, line 739: and here

graphblas/vector.py
```diff
@@ -725,13 +736,20 @@ def apply(self, op, right=None, *, left=None):
...
+                op = get_typed_op(
+                    op, self.dtype, right.dtype, is_right_scalar=True, kind="indexunary"
+                )
+                cfunc_method = "IndexOp"
+            else:
+                op = get_typed_op(op, self.dtype, right.dtype, is_right_scalar=True, kind="binary")
+                cfunc_method = "BinaryOp2nd"
             if op.opclass == "Monoid":
```
eriknw, line 747: and here (indented)

eriknw (commented):
w00t!  Thanks Jim!  I had mostly (entirely?) minor nits.  You know, it's already been super-nice having select operators.  I forget where IndexUnary operators will be used.  Less often, though, right?  It was a good call to separate them.

I'm fine taking an incremental approach to PRs.  One should implement what one is comfortable with, and hopefully anything still to be done is noted somewhere or intended to be implemented soon-ish.

Btw, a speaker at PyCon gave an impassioned speech about about reaching and *staying at* 100% coverage as a virtue and making your project better.  So, this is a friendly reminder to please look at all missed coverage at https://coveralls.io/github/python-graphblas/python-graphblas.  We're not sticklers about coverage (well, I kinda am for myself), but good coverage is appreciated.  I think I may re-enable the coveralls chat bot.  I suppose it wasn't _that_ noisy, I just wasn't used to it is all.

graphblas/operator.py
```diff
@@ -2764,6 +2840,7 @@ def op_from_string(string):
         semiring_from_string,
         # Give binary precedence over this
         select_from_string,
+        indexunary_from_string,
```
eriknw, line 2843: What would one expect to get from e.g. `op.from_string("tril")`?  Should we look for select before indexunary, b/c select is more common?

And, remind me, when we register a new UDF, do we plan to create both versions, or just the opclass that is being registered?

eriknw:
And sorry about the failed tests.  I was stupid.  Merge with main for a fix.

graphblas/operator.py
jim22k: Yes, we should look for select before indexunary. If it exists in select, that is most likely what the user wanted. And if you do `op.from_string("rowindex")`, you will get the indexunary version because it doesn't exist in select. So anything returning bool will default to select while everything else will be apply.

For UDFs, I haven't decided yet what to do. My initial thought is to register all places where it makes sense. So always register the indexunary, and additionally register with select if it returns a bool.

jim22k:
As far as coverage goes, I know I'm missing several lines of code which get tested once I add UDFs. Until then, I don't want to remove that code or I might forget to add it back in.

eriknw (approved):
🚀

PR #235, opened 2022-05-17
Have `op.from_string` include aggregators (last priority)

eriknw:
Although aggregators aren't present in `gb.op`, `gb.op.from_string` often serves to "convert this string to the best operator-like thing".


jim22k (approved):
As long as these remain at the bottom of the list, this feels okay to me.

PR #236, opened 2022-05-17
Deprecate `require_monoid=` keyword

eriknw:
@jim22k you know my thoughts on `require_monoid=`.  In practice, it's super-duper annoying.  We proceeded conservatively: it's easier to relax than constrict, and I think it's now time to relax this.

More code cleanup will be possible once we actually remove it.

### DIFF
--- a/graphblas/tests/test_matrix.py
+++ b/graphblas/tests/test_matrix.py
@@ -384,17 +384,17 @@ def test_ewise_add(A):
         [2, 0, 1, 2, 2, 2, 3, 3, 4, 4, 5, 5, 6],
         [4, 3, 5, 3, 8, 5, 3, 7, 8, 3, 1, 7, 4],
     )
-    with pytest.raises(TypeError, match="require_monoid"):
-        A.ewise_add(B, binary.second)
+    # with pytest.raises(TypeError, match="require_monoid"):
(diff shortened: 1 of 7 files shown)


graphblas/tests/test_matrix.py
```diff
@@ -384,17 +384,17 @@ def test_ewise_add(A):
...
         [4, 3, 5, 3, 8, 5, 3, 7, 8, 3, 1, 7, 4],
     )
-    with pytest.raises(TypeError, match="require_monoid"):
-        A.ewise_add(B, binary.second)
+    # with pytest.raises(TypeError, match="require_monoid"):
+    A.ewise_add(B, binary.second)  # okay now
     # surprising that SECOND(x, empty) == x, which is why user
     # must opt-in to using binary ops in ewise_add
-    C = A.ewise_add(B, binary.second, require_monoid=False).new()
+    C = A.ewise_add(B, binary.second).new()
```
jim22k, line 391: Update comments. The opt-in requirement no longer exists.

jim22k (approved):

graphblas/tests/test_matrix.py
eriknw: Done.  Thanks for the attention to detail!

PR #237, opened 2022-05-18
Add a few missing numpy ufuncs: cbrt, float_power, positive

eriknw:
Note that `cbrt` is coming in SuiteSparse:GraphBLAS 7.1.0 🎉 .

This came about b/c I noticed `cbrt` (the principal cube root; e.g., `cbrt(-8) == -2`) was missing when I wanted to use it in `graphblas-algorithms` for weighted clustering coefficients.

Also, I noticed the numba documentation we were pointing to is out of date.

### DIFF
--- a/graphblas/binary/numpy.py
+++ b/graphblas/binary/numpy.py
@@ -2,7 +2,7 @@
 
 See list of numpy ufuncs supported by numpy here:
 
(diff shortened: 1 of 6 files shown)


eriknw:
And this is why we have a test matrix 😞

jim22k (approved):

PR #243, opened 2022-05-22
Allow `A << 1` to create dense matrices.

eriknw:
Now that SuiteSparse:GraphBLAS has iso-valued objects, I think it's okay to support this syntax.
Also, when reading or writing algorithms, I find that type-ambiguity isn't much of an issue.
So, I don't think the nuisance of raising this exception is worth the benefit anymore.

### DIFF
--- a/graphblas/base.py
+++ b/graphblas/base.py
@@ -396,19 +396,7 @@ class BaseType:
                                 "AmbiguousAssignOrExtract, and scalars."
                             ) from None
                     updater = self(mask=mask, accum=accum, replace=replace, input_mask=input_mask)
-                    if type(self) is Matrix:
-                        if mask is None:
-                            raise TypeError(
-                                "Warning: updating a Matrix with a scalar without a mask will "
-                                "make the Matrix dense.  This may use a lot of memory and probably "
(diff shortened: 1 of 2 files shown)


jim22k:
What does happen if we run out of memory? I'm thinking of Jeremy Kepner's frequent comments that we can't ever realize a hypersparse matrix or we blow up. If someone were to accidentally type C << 1 with a 2**60x2**60 hypersparse, what would happen?
1. Would the C code give a nice error message?
2. Would the Python process freeze and/or die?
3. Would the computer completely lock up?

As long as the behavior is (1), we're good. If it's (2), we should think about this and possibly guard against hypersparse explosions. If it's (3), then we should not change our syntax.

eriknw:
The result is (1) for SuiteSparse:GraphBLAS: we get a good error message originating from C if we run out of memory.

But, doing this in SuiteSparse:GraphBLAS is fine due to iso-valued objects:
```python
C = Matrix(int, 2**60, 2**60)
C << 1
```
and this is what will raise `OutOfMemory`:
```python
del C[0, 0]
```

eriknw:
YOLO.  Merging 🚀

Issue #234, opened 2022-05-17
Advanced Notation


jim22k:
These were left over from discussions in #147 

1. Implement more flexible masks
   - Main need is allowing boolean objects as masks without explict .V suffix
2. Revisit A[A<1] notation
3. Revisit complex set expressions ((A > 1) & (A < 7), A.S | B.S, etc)
4. Revisit using masks in select (as a convenience recipe) (e.g. A.select(B.S))

eriknw:
I think number (1) (using boolean objects as value masks) will give us the biggest bang for the buck.  It should be easy to implement and handy to use.  It also unlocks some of the other notation.

Regarding (4), would the recipe for `updater << A.select(B.S)` be:
  - `updater(mask=B.S) << A` if updater doesn't have a mask
  - `updater << A.dup(mask=B.S)` if updater does have a mask

I guess (1) and (4) combined would allow e.g. `A.select(B < 5)`.  I don't hate this (it may actually be kinda nice), and it makes it easier to punt on `A[A < 1]` and `A[B < 1]` notation, which might be a can of worms.

(3) already works with boolean objects, `(A > 1) & (A < 7)` etc.  The thing that is missing is treating masks as value-y.  The first step to do so is to give masks `.new(dtype=None, *, mask=None, name=None)`.  We can add `.new` to masks and still decide to not treat them as value-y.  This exercise can help us figure out how to best combine masks.  Possible recipes for `M1.new(mask=M2)` are:
  - One structural mask `S`, any other mask `A`:
    - `unary.one(S.parent).new(mask=A)`
  - One value mask `V`, any other mask `A`:
    - `select.nonzero[bool](V.parent).new(mask=A)`
  - Both complemented masks, `bigger` and `smaller` where `bigger.nvals >= smaller.nvals`:
    - `result = Matrix(...)`
    - `result(bigger) << True`
    - `r(smaller, replace=True) << result`
    - Realizing complemented masks can be very expensive; let's hope it's not done for hypersparse objects!
    - Alternative to explore:
      - `temp = binop(bigger.parent | smaller.parent).new()`
      - `result(~temp.X) << True`
      - where `binop` and `X` are chosen as appropriate (such as `pair`-`S` for S-S, `lor`-`V` for V-V)
      - Still need to figure out when one mask is value and one is structural
        - In the cases where it works, perhaps we can decide whether this or the previous recipe will be better based on nvals.

[...]

PR #248, opened 2022-05-25
Allow boolean objects to be used as value masks

eriknw:
Was playing around a bit tonight and came up with this.  Sorry for taking "easier" work if anybody else had designs to implement this!

This is (1) in https://github.com/python-graphblas/python-graphblas/issues/234

An interesting side effect of applying our normal auto-compute rules is this gave `.S` and `.V` to `TransposedMatrix` objects, which required `config.get("autocompute")` to be true.  But, we don't need to add this.  Thoughts?  It's late, and I can't think.

Also, I'm not 100% sold on this PR, but I think it will be handy during the hands-on exploration phase.  I wonder whether this syntax will be used in "hardened" algorithms, or whether there will usually or always be faster recipes (that may not be obvious or as conveniently written) in practice.  I bet there are simple examples where this syntax is "the best", but, yeah, late, can't think.

### DIFF
--- a/graphblas/base.py
+++ b/graphblas/base.py
@@ -3,6 +3,7 @@ from contextvars import ContextVar
 from . import config, ffi
 from . import replace as replace_singleton
 from .descriptor import lookup as descriptor_lookup
+from .dtypes import BOOL
 from .exceptions import check_status
 from .expr import AmbiguousAssignOrExtract, Updater
 from .mask import Mask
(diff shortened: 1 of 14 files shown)


eriknw:
@jim22k and I decided to _not_ allow TransposedMatrix objects to be used as masks, and this PR has been updated.  There isn't a strong argument to include it, and the solution (e.g., `A.T` -> `A.T.new()`) is straightforward and short.

eriknw:
This is going in!

PR #250, opened 2022-05-30
Assign list to many indices, such as `v[[1, 2]] = [3, 4]`

### LINKED ISSUE #245: Assign a list to many indices, such as `v[[1, 2]] = [3, 4]`
I don't know how difficult this would be to support, but it would be nice to support things like the following:
- `v[[1, 2]] = [3, 4]`
- `v[:] = np.arange(v.size)`

Currently (for Vectors), if the right-hand-side isn't a Vector, it is assumed to be a Scalar.  We have this behavior, b/c this matches the C API.  There is no C API for assigning a C array to indices.

For SuiteSparse:GraphBLAS, the efficient thing to do would be to use `ss.import_full` to create the Vector or Matrix.  It would be easiest and fastest if no value (such as `None`) in the list or array could mean "missing" so that one can always use `ss.import_full`.

For subarray UDTs, there is the possibility of ambiguity.  Does `v[indices] = subarray` mean to assign `subarray` value to all elements (as would currently be done), or to broadcast and assign element-by-element (as proposed in this issue).  This is a very narrow edge case: it needs to be a subarray UDT where the subarray and the index have the same size (although this begs the question whether we want to use the sizes to determine the intent).  The easiest solution is to assume "scalar" assignment when using subarray UDTs and require users to create Vector or Matrix objects on the RHS if they want to assign element-wise.

eriknw:
Closes #245.  This also fixes some import functions for sub-array UDTs.
Shapes must match exactly.  We aren't forgiving or do broadcasing like NumPy.

I agree in principle that *if* we were to allow operations between
e.g. vectors and numpy arrays or lists (such as `binary.min(v & a)`),
then we should probably be consistent with how we treat
lists/arrays/etc as vectors or matrices when assigning.
I think we should treat them as dense in all cases.

I don't think using a list to assign multiple items at a time *implies*
that a list or array should be treated as vectors or matrices elsewhere.
The mental model is that it's assigning multiple elements.  That's it.
That we convert it to a Vector or Matrix is an implementation detail.

I think assigning lists/arrays is intuitive, and useful in an interactive,
educational setting.  It's plausible that it will be useful in practice too
to conveniently and efficiently assign arrays to sub-vector/sub-matrices.

Adding this functionality was pretty straightforward and very isolated.
Most of the code is to enable good error messages and handling sub-array UDTs.

### DIFF
--- a/graphblas/_ss/matrix.py
+++ b/graphblas/_ss/matrix.py
@@ -2164,7 +2164,7 @@ class ss:
         if bitmap is values:
             values = np.copy(values)
         if method == "import":
-            nrows, ncols = get_shape(nrows, ncols, values=values, bitmap=bitmap)
+            nrows, ncols = get_shape(nrows, ncols, dtype, bitmap=bitmap, values=values)
         else:
             nrows, ncols = matrix.shape
         Ab = ffi_new("int8_t**", ffi.from_buffer("int8_t*", bitmap))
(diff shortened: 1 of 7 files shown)


eriknw:
Thanks for the review and careful consideration @jim22k!

PR #251, opened 2022-05-31
Recipe to perform select with a mask (or bool) argument.

eriknw:
This allows e.g. `A.select(B.S)` and `A.select(C < 5)`.

This is enhancement (4) in #234.

This does not yet optimize e.g. `A.select(A < 4)` into the suitable select operation, but such an optimization is possible (and would probably be straightforward-ish).

Also, we'll want to double-check the recipes regarding masks, accum, etc.  I think maybe if there is no mask but there is an accumulator, then we shouldn't use `replace=True`.

### DIFF
--- a/graphblas/matrix.py
+++ b/graphblas/matrix.py
@@ -5,12 +5,12 @@ import numpy as np
 
 from . import _automethods, backend, binary, ffi, lib, monoid, select, semiring, utils
 from ._ss.matrix import ss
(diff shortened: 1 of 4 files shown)


jim22k (approved):
Looks good

eriknw:
Thanks for the review (and suggestion) Jim, I think this reads pretty nicely.

I'm finding our custom-expression handling is actually pretty powerful.

Merging!

PR #254, opened 2022-06-02
Add SuiteSparse:GraphBLAS (de)serialize

### LINKED ISSUE #200: Add SuiteSparse:GraphBLAS (de)serialization methods
`GxB_Matrix_serialize` and `GxB_Vector_serialize` (and deserialize).

I'm not sure whether we should use these for pickling.  The current method to pickle uses `x.ss.export()`, and then a system like Dask can choose to compress the resulting arrays if it determines doing so is worthwhile.  Until we can measure communication performance in Dask's distributed scheduler, perhaps we could add a configuration for which method to use to pickle.

We could still choose to expose (de)serialize as e.g. `x.ss.serialize()`.

eriknw:
Fixes #200.

The GxB version of these methods is nicer to work with.  This PR still needs docs and probably a little more cleanup and testing.

Is it okay if `x.ss.serialize()` returns a numpy array of uint8, or would bytes be better?  It's easy to do `val.tobytes()`, but this performs a data copy, so I prefer to return a numpy array.

`.ss.deserialize` accepts a numpy array or buffer.

### DIFF
--- a/graphblas/_ss/descriptor.py
+++ b/graphblas/_ss/descriptor.py
@@ -1,0 +1,78 @@
+from graphblas import ffi, lib
+from graphblas.descriptor import Descriptor, _desc_map
+from graphblas.exceptions import check_status
+
+str_to_compression = {
+    "none": lib.GxB_COMPRESSION_NONE,
+    "default": lib.GxB_COMPRESSION_DEFAULT,
+    "lz4": lib.GxB_COMPRESSION_LZ4,
+    "lz4hc": lib.GxB_COMPRESSION_LZ4HC,
+}
+
+
+def get_nthreads_descriptor(nthreads, _cache=True):
+    nthreads = max(0, int(nthreads))
+    key = ("nthreads", nthreads)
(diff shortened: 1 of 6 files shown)


jim22k (approved):

jim22k:
numpy array of uint8 should be fine. With serialize/deserialize, speed is the main concern, so making things a little less intuitive to achieve the fastest roundtrip feels justified.

eriknw:
Alright, I updated this to support (de)serializing of UDTs.  We use the name (which has finite-width) to save the numpy repr from which we can recreate the dtype using `eval`.  This pretty much "just works", but there are some sub-optimal behaviors:
- The name can be long and ugly, and may not be understood by other libraries or languages
  - If the UDT was given a name (and it's not anonymous), should we always use it? (this has other challenges)
- Recreating the dtype requires using `eval`.  I'm not sure how to get around this.
  - The _user_ can get around this by providing `dtype` to `deserialize`.

I'd like to move forward as-is to see if it's workable, and we can update these strategies if/when needed.

PR #256, opened 2022-06-03
Safely deserialize UDT dtypes

eriknw:
It's surprisingly annoying to convert a numpy dtype to and from a string, but this seems to do the trick and let's us do it safely without `eval`.


eriknw:
Upon closer reading of the documentation for SuiteSparse:GraphBLAS, the name of the dtype should be a valid C identifier.  This is to ensure future compatibility for when a JIT is added.

I'll figure out what to do about the name thing in the future.  Merging this PR, which is an improvement.

PR #257, opened 2022-06-04
Apply with a dict 🎉 

### LINKED ISSUE #246: Apply dict to map values
This is probably a crazy idea (and I don't currently have a use case), but could we make it easy to apply a dict to map values?  For example, to make the following easier and more reliable (i.e., not rely on inlining the dict):
```python
In [10]: v = gb.Vector.from_values(np.arange(5), np.arange(5))

In [11]: v
Out[11]:
"v_2"      nvals  size  dtype  format
gb.Vector      5     5  INT64    full
-------------------------------------
index  0  1  2  3  4
value  0  1  2  3  4

In [12]: v.apply(lambda x: {0: 0, 1: 2, 2: 4, 3: 6, 4: 8}[x]).new()
Out[12]:
"v_4"      nvals  size  dtype  format
gb.Vector      5     5  INT64    full
-------------------------------------
index  0  1  2  3  4
value  0  2  4  6  8
```
This requires Numba to know the types of the dictionary.  Also, I don't know how easy it would be to have a default value.  This may require its own method.

This is inspired by the Pandas functionality to be able to use `x.map(d)` to apply a dict or Series mapping.

eriknw:
Closes #246.

`A.apply({1: 10, 2: 20}, 999)`

Heh, I know this isn't high priority (at all!), but it is oh-so-fun and was quick to do.  I think the function we jit with numba is pretty competitive.

I doubt this handles UDTs, and we could probably be a bit smarter with dtypes in general.

### DIFF
--- a/graphblas/matrix.py
+++ b/graphblas/matrix.py
@@ -1,5 +1,6 @@
 import itertools
 import warnings
+from collections.abc import Mapping
 
 import numpy as np
 
@@ -10,7 +11,14 @@ from .dtypes import _INDEX, FP64, lookup_dtype, unify
 from .exceptions import DimensionMismatch, NoValue, check_status
 from .expr import AmbiguousAssignOrExtract, IndexerResolver, Updater
 from .mask import Mask, StructuralMask, ValueMask
-from .operator import UNKNOWN_OPCLASS, find_opclass, get_semiring, get_typed_op, op_from_string
+from .operator import (
+    UNKNOWN_OPCLASS,
+    _dict_to_func,
+    find_opclass,
+    get_semiring,
+    get_typed_op,
+    op_from_string,
+)
 from .scalar import (
     _MATERIALIZE,
     Scalar,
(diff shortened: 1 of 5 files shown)


jim22k:
I'm currently -1 on this. The mental model I see here is that we can take a dict and construct a Matrix from it, then use that to update another Matrix. In that case, it would be assign, not apply.

Looking at your example was not at all obvious what was going on. And that is the worry I have with each new addition. They seem simple, but unless someone knows all the corners and shortcuts, it makes it harder to read someone else's code.

I'm open to being convinced, but these are my initial thoughts on the proposal.

eriknw:
Round(s) of convincing (and likely improvements) is expected and appreciated.

> The mental model I see here is that we can take a dict and construct a Matrix from it, then use that to update another Matrix. In that case, it would be assign, not apply.

I have no idea what you mean by this.  The equivalent recipe is:
```python
C = Matrix(A.dtype, A.nrows, A.ncols)
C(A.S) << default
for k, v in d.items()
    C(A == k) << v
```
This is just like `df.map(d)` in pandas.  `map` would probably be a better method name for us too--it reads better, and `apply` is already pretty loaded.  Having a separate method also make future enhancements easier (no idea what, but I'm regularly impressed by the savviness of users and their suggestions).

I don't doubt use cases for this will arise (supposing we get more users), and alternatives to perform this are pretty ugly and most likely slower.  So, I think this is worth exploring, but we can keep it on the back-burner and wait until we have conventions for showing examples in docstrings.  We've been pushing core features forward faster than we can document them (with good reason), but we need documentation to catch up, and we should be in the habit of requiring sufficient documentation for PRs like this.

Possible enhancements:
- It would be nice if we could come up with an efficient recipe that could result in missing values instead of using a default value.
- Alternatively, to use the original value instead of default or missing.
- Maybe get the default value from `d.__missing__` if available (pandas does this).

eriknw:
Okay, I think the natural API for this is:
```python
def map(self, mapping, default=None):
```
where `mapping` is a dict-like `Mapping`.  This returns an *expression* and does not modify the original.

`default` can be:
- `None`: drop non-matching values
- scalar: use this as the default value
- `Matrix`: get default values from this object
  - if default is self, then the original value is "passed through", and we can have a more efficient recipe
  - `mymatrix.map` accepts `Matrix` and `myvector.map` accepts `Vector` (naturally)

I think the keys should be coerced to the dtype of the parent object dtype, and values coerced to the output object dtype.  This will let us handle UDTs and awkward dtypes like `uint64`.

Issue #20, opened 2020-05-11
Add implementation-specific descriptors


eriknw:
SuiteSparse:GraphBLAS includes many extensions to the official GraphBLAS specification, and one can reasonably expect other implementations (a.k.a. backends) to offer their own (possibly different) additions.  We should find a non-intrusive yet clear way to support these.

When creating new objects, we can accept `**backend_kwargs` where valid keywords are determined and handled by the specific backend.  For example, when creating a new matrix, we may have `ss_format='by_row'`.  In this example, we have all SuiteSparse extension keywords begin with `ss_`.

Similarly, backends may want to have additional methods or properties.  For SuiteSparse, this may look like `A.ss.select(...)`, `A.ss.format`, `A.ss.pattern`, etc.

eriknw:
@jim22k discussed this briefly today, and we think global and object-specific options should probably be MutableMapping on `.ss` namespaces, such as `A.ss.config` (object-specific) and `gb.ss.config` (global) (or `ss.options`).  `config.__doc__`/`config?`/`help(config)` should describe the options and list valid values for each option.  Setting values should normal Python objects such as int, float, or str, which we _may_ need to map to a C enum.  When creating new objects, perhaps a mapping could be provided via e.g. `**kwargs` or `config={...}` to set the config options.

It _may_ be nicest to specify additional descriptors (such as AxB_method, nthreads, chunk, gpu_control, etc.) with a context manager.

eriknw:
The global SuiteSparse config--`gb.ss.config`--added in https://github.com/python-graphblas/python-graphblas/pull/258 is able to globally set _some_ of the SuiteSparse descriptor fields, such as nthreads and chunk, which is great, but doesn't completely solve (i.e., expose) everything.

Here's an idea.  Let's allow passing an `Options` object on the left-hand side.  For example
```python
A(Options(nthreads=4, AxB_method="gustavson", sort=False)) << B @ C
```
Two nice things about an object is it has clear intent and can be reused, such as
```python
opts = Options(nthreads=4, AxB_method="gustavson", sort=False)
A(opts) << B @ C
```
I think it would be sufficient to allow this only when updating an existing object.  That is, I don't think we need to add `options=None` or `**kwargs` to every `.new` method (but we could if/when there is sufficient need, and I would prefer `options=` over `**kwargs`).

`Options` would need to be backend-specific, and a vanilla implementation has no available options.

[...]

PR #258, opened 2022-06-05
Add `ss.config` to manage SuiteSparse:GraphBLAS config options

eriknw:
Some of the code is a little hairy, but I think it gives a pretty good user experience.


eriknw:
This addresses one of the primary things from SuiteSparse:GraphBLAS that we were missing.  xref: https://github.com/python-graphblas/python-graphblas/issues/20#issuecomment-1143825364

jim22k (approved):

PR #260, opened 2022-06-06
`{to,from}_scipy_sparse` array

### LINKED ISSUE #176: PERF: make `gb.io.from_scipy_sparse_matrix` faster
xref: https://github.com/GraphBLAS/python-suitesparse-graphblas/issues/34

Our data conversions in `grblas.io` could use some love in general.  But, an easy optimization we could do is to use e.g. `Matrix.ss.import_csr` in `grblas.io.from_scipy_sparse_matrix` instead of always converting the scipy matrix to COO.

@steffen-sch, any interest?

### LINKED ISSUE #181: Use scipy sparse array, not matrix
`grblas` tests raise this warning from `scipy.sparse`:
```
grblas/tests/test_io.py::test_matrix_to_from_networkx
  /home/erik/git/grblas/grblas/io.py:38: DeprecationWarning:

  The scipy.sparse array containers will be used instead of matrices
  in Networkx 3.0. Use `to_scipy_sparse_array` instead.
    ss = nx.convert_matrix.to_scipy_sparse_matrix(g)
```

eriknw:
This closes #176 and closes #181.

Do we like the names `from_scipy_sparse` and `to_scipy_sparse`?  We call others `from_numpy` and `from_networkx`, not `from_numpy_array` or `from_networkx_graph`, so I think it's fine to not specifically call out `_array`.  Also, `from_scipy_sparse` handles both array and matrix.

We called the old functions `from_scipy_sparse_matrix` originally, because we knew that scipy matrix was going to be deprecated, so we wanted to call it out and have the most flexibility when scipy sparse arrays finally came.


graphblas/io.py
```diff
@@ -183,15 +287,8 @@ def mmwrite(target, matrix, *, comment="", field=None, precision=None, symmetry=
...
     except ImportError:  # pragma: no cover
         raise ImportError("scipy is required to write Matrix Market files") from None
-
-    # Note: it would be preferable to write the data in full format if it is indeed full.
-    #       However, there may be a bug: https://github.com/scipy/scipy/issues/13634
-    #
-    # SS, SuiteSparse-specific: format
-    # fmt = matrix.ss.format
-    # if fmt == 'fullr' or fmt == 'fullc':
-    #    array = matrix.ss.export(format=fmt)['values']
```
eriknw, line 193: btw, I uncommented these lines and changed the minimum required version of scipy to 1.7.0, which is the first release with the fix for the issue in the code comment above.

jim22k (approved):

PR #267, opened 2022-07-04
Allow masks to be combined with `|` and`&`

### LINKED ISSUE #234: Advanced Notation

eriknw:
Closes #234

These recipes are much faster than the equivalent recipes in the docstrings, but they still aren't hyper-optimized (they're only *semi*-optimized).  As with `.new`, this is structured so that any recipe can be updated if need arises.

Currently, all recipes return structural masks that may be complemented.  It's not required for them to be structural--perhaps it would be better if some were value masks.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@
 # To run: `pre-commit run --all-files`
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.2.0
+    rev: v4.3.0
     hooks:
       - id: check-added-large-files
       - id: check-yaml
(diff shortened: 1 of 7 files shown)


jim22k (approved):
The code all looks good, but the comments are technically wrong.

I think the recipe needs to be:
```
>>> val = Matrix(bool, nrows, ncols)
>>> val(mask1) << True
>>> val(mask2, replace=True) << val
>>> new_mask = val.S

The same recipe should be listed in the comments for `.new()` as well.
```

eriknw:
Updated docstrings.  We came up with the same solution, which is a good sign.

PR #268, opened 2022-07-06
[WIP] Fill out the documentation

jim22k:
This uses sphinx with the `pydata_sphinx_theme` for styling.

The goal for this first cut is correct documentation covering most of the library to get users started. More work is expected in the future to cover all the corner cases, but we can't let perfection become the enemy of progress.


eriknw:
yay :tada:

I love this so much.

I'm not sure why the style checks fail on CI but pass pre-commit.  Perhaps we should simply run pre-commit in CI.  I have a fix for the flake8 errors in this PR: https://github.com/python-graphblas/python-graphblas/pull/267

PR #274, opened 2022-08-04
Add API Reference docs

jim22k:
Still not a proper logo, but looks better than all flat text


eriknw:
LGTM, thanks @jim22k!  Could the bounding box be updated?  Looks like there's still a gap on the right.

jim22k:
Hopefully fixed the SVG logo offset issue.

...and I started looking at how hard it would be to autodoc the API reference guide. I'm now more than halfway done with updating the docstrings to make the API reference guide functional.

eriknw (approved):
Love this!  Feel free to merge as you please.

graphblas/matrix.py
```diff
@@ -136,16 +141,39 @@ def _deserialize(pieces, name):
...
+        .. code-block:: python
+
+            del M[1, 5]
+        """
         del Updater(self)[keys]
 
     def __getitem__(self, keys):
+        """
+        Extract a single element, row/column, or submatrix.
+        The `user guide <../user_guide/operations.html#extract>`__ contains more details.
```
eriknw, line 169: We really ought to try to follow PEP 257 and have a one liner at the beginning followed by a newline:

https://peps.python.org/pep-0257/

graphblas/matrix.py
```diff
@@ -136,16 +141,39 @@ def _deserialize(pieces, name):
...
+        """
         return ValueMask(self)
 
     def __delitem__(self, keys):
+        """
+        Delete a single element, row/column, or submatrix.
+
+        .. code-block:: python
+
+            del M[1, 5]
```
eriknw, line 162: Can we instead use `>>>` for code blocks?

jim22k: Most of what I'm showing are complementary one-line statements (method vs functional) with no output.

I personally like this look better.
CurrentNotation

This just feels awkward and must be split into two separate code blocks because comments don't make sense when using `>>>` command-prompt style.
UseTripleLT

eriknw: Docstrings aren't just read from the website though.  Reading via a REPL or the file is also very common.

An argument for the RST version is that it is friendlier to copy/paste wherever it is.

eriknw: So we can use `.. code-block:: python` in docstrings if you'd like.  It does look better when rendered.

graphblas/operator.py
```diff
@@ -1532,13 +1595,31 @@ def _from_indexunary(cls, iop):
...
     def register_new(cls, name, func, *, parameterized=False, is_udt=False, lazy=False):
+        """
+        Register a SelectOp. The name will be used to identify the SelectOp in the
+        ``graphblas.select`` namespace.
+
+        The function will also be registered as a IndexUnaryOp with the same name.
+
+        .. code-block:: python
+
+            >>> from graphblas.operator import SelectOp
```
eriknw, line 1618: This is interesting though.  Do we need both `.. code-block:: python` and `>>>`?

graphblas/operator.py
```diff
@@ -2379,12 +2496,31 @@ def _compile_udt(self, dtype, dtype2):
...
     @classmethod
     def register_anonymous(cls, binaryop, identity, name=None):
+        """
+        Register a Monoid without registering it in the ``graphblas.monoid`` namespace.
+
+        Because it is not registered in the namespace, the name is optional.
+
+        :param BinaryOp binaryop: A builtin or registered binary operator
+        :param identity: The identity value of the monoid
+        :param str name: The name associated with the monoid
```
eriknw, line 2506: In general, I prefer to mimic what other PyData libraries do such as pandas, dask, xarray, etc.  I rarely see RST in docstrings such as `:param`.  Does the typical way render poorly? e.g.
```
Parameters
----------
binaryop : BinaryOp
```
sphinx+pandoc are pretty flexible with this, aren't they?

graphblas/operator.py
jim22k: Just tried it out and `>>>` alone will work. No need for additional code-block notation.

graphblas/operator.py
jim22k: This style was not working for me, but now that I'm looking at other projects, I think I need to include the `numpydoc` sphinx extension. I agree that the numpy style is easier to read as plain text.

eriknw: Good to know, and thanks!  I really should get more familiar with creating docs...

graphblas/matrix.py
jim22k: Interesting that [numpy prefers PEP 257 style](https://numpydoc.readthedocs.io/en/latest/format.html) while [pandas prefers the style I have been writing](https://pandas.pydata.org/docs/development/contributing_docstring.html).

I'm fine switching to the numpy style.

eriknw: Pandas style is great, and it still prefers a single line short summary at the beginning followed by empty line:
https://pandas.pydata.org/docs/development/contributing_docstring.html#section-1-short-summary

graphblas/operator.py
jim22k: Okay, I got it working for one method. The rendered style is slightly different, but still acceptable. I'll move everything which is autodoc'd over to the numpy parameter style.

eriknw: Awesome, thanks!

We could also consider the google style:
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/index.html#google-vs-numpy
but I think likely users/contributed will be more familiar with numpy-style.  I'm open to either.

eriknw:
If you're bold, you could try to see what `pydocstyle` complains about:
http://www.pydocstyle.org/en/stable/

If we make `pydocstyle` happy, we could add it to our linters.

eriknw:
We could also give `velin` a try, but I would make sure all changes are saved first!
https://github.com/Carreau/velin

Super-happy to have better documentation and docstrings, thanks Jim!

jim22k:
Merging now. We can keep fiddling with the style in the future, but it's good enough for now.

Issue #253, opened 2022-06-01
"from components" to encapsulate common `to_coo()` / `from_coo()` recipes


eriknw:
_**Edit:** note that `to_values` is deprecated; use `to_coo` instead_

We have found that we sometimes perform `to_values()` followed by `from_values()` to perform different operations such as creating a diagonal matrix or using the values as an index.

We may want to give this common(-ish) pattern a name.  For example:
```python
Matrix:
    @classmethod
    def from_components(
        row_obj,
        col_obj,
        val_obj,
        dtype=None,
        *,
        row_how="row",
        col_how="col",
        val_how="val",
        dup_op=None,
        nrows=None,
        ncols=None,
        name=None,
    ):

Vector:
    @classmethod
    def from_components(
        index_obj,
        val_obj,
        dtype=None,
        *,
        index_how="index",
        dup_op=None,
        size=None,
        name=None,
    ):
```
For example, in `generalized_degrees` https://github.com/python-graphblas/graphblas-algorithms/pull/11, we compute a number for each edge, and then we calculate a histogram of counts using the column index as the count:
```python
    rows, cols, vals = Tri.to_values()
    # The column index indicates the number of triangles an edge participates in.
    # The largest this can be is `A.ncols - 1`.  Values is count of edges.
    return Matrix.from_values(
        rows,
        vals,
        np.ones(vals.size, dtype=int),
        dup_op=binary.plus,
        nrows=A.nrows,
        ncols=A.ncols - 1,
        name="generalized_degree",
    )
```
would be written as
```python
    return Matrix.from_components(
        Tri,
        Tri,
        1,
        col_how="val",
        dup_op=binary.plus,
        ncols=Tri.ncols - 1,
        name="generalized_degree",
    )
```
`row_how`, `col_how`, and `val_how` should be one of `{"index", "row", "col", "val"}`.  For Vectors, we treat `"row"` and `"col"` the same as `"index"`.  It is an error to use `"index"` for Matrix objects.

Default `nrows` and `ncols` are inferred from what they come from.  For example, for `nrows`:
- `row_how == "row"`: `nrows = row_obj.nrows`
- `row_how == "col"`: `nrows = row_obj.ncols`
- `row_how == "val"`: `nrows = max(row_obj) + 1`

[...]

PR #276, opened 2022-08-07
Explore `to_values` that doesn't return everything.

eriknw:
Lots of ways to spell this.  Pretty low priority.

I wanted this for myself so I could experiment with it.


eriknw:
Performance-wise, this optimization is definitely worth it if you only want one of the components of a matrix, or two components including values.  If you only want values, it's super-duper fast.

We'll want this functionality if we ever implement https://github.com/python-graphblas/python-graphblas/issues/253.

I'm not crazy about the API in this PR.  Suggestions?

PR #279, opened 2022-09-12
WIP: Begin adding viz functions for viewing sparse matrix

eriknw:
One uses matplotlib `spy` and one uses datashader via hvplot.  Both are pretty!

These are meant to be opinionated and easy to use with reasonable defaults, but with "enough" flexibility to allow for some customization.  They're also meant to show off :)

matplotlib:
spy1

datashader via hvplot:
bokeh_plot

Some nice things about using datashader:
- interactive!
- able to use different aggregators
- able to view the _values_
- able to compose and link multiple plots together
  -  could be useful for e.g. viewing both arguments to a function

### DIFF
--- a/graphblas/__init__.py
+++ b/graphblas/__init__.py
@@ -74,6 +74,7 @@ _SPECIAL_ATTRS = {
     "unary",
     "utils",
     "vector",
+    "viz",
 }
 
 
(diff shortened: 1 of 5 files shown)


jim22k:
Looks nice!

eriknw:
Merging shortly so we can do a release.

These functions deserve better docstrings, including images to show on the website.

Also, they probably deserve a notebook, and existing notebooks need updated to use `graphblas.viz` instead of `graphblas.io`.

But, I think these are probably good enough to ship 🚀

PR #280, opened 2022-09-23
Connection to Awkward Arrays

jim22k:
Awkward Arrays feel very similar to sparse Matrices, except that all values are pushed to one side rather than living at different column indices.

To enable experimentation between these two libraries, this PR builds opinionated connectors from GraphBLAS objects to Awkward Arrays.


eriknw (commented):
I'm not very familiar with using `awkward`, so it's difficult for me to comment or review whether this is a good way to connect to it.  In particular, is this the most natural way to _come from_ Awkward?  Are there any argument we can/should expose to make it more flexible?  What if there are no indices--could they be "compatified" to e.g. the left?  Does Awkward have a concept of length of a ragged dimension (such as the longest row)?

Also, you have uncovered lines in this PR.  You have uncovered lines from previous PRs too.  Can you please take a look at https://coveralls.io/github/python-graphblas/python-graphblas ?  Thanks.

Otherwise, this looks pretty good as far as I can tell.

eriknw:
I also wonder whether we should add an `"auto"` option for the format in some of the `to_` functions.  For example, why default to `"csr"` if the original matrix is `"csc"`?

graphblas/io.py
```diff
@@ -319,6 +375,120 @@ def to_scipy_sparse(A, format="csr"):
...
+        )
+        d = {
+            "node1-offsets": info["indptr"],
+            "node3-data": info[indices],
+            "node4-data": info["values"],
+        }
+        if format.startswith("hyper"):
+            # Define behaviors to make all fields function at the top-level
+            @ak.behaviors.mixins.mixin_class(ak.behavior)
+            class DoublyCompressedMatrix:
```
eriknw, line 461: Want to make this a global that is only defined once?  e.g.
```python

_DoublyCompressedMatrix = None

def to_awkward(A, format=None):
    ...
    global _DoublyCompressedMatrix
    if _DoublyCompressedMatrix is None:
        @ak.behaviors.mixins.mixin_class(ak.behavior)
        class _DoublyCompressedMatrix:
            ...
```

eriknw (approved):
Thanks for responding to my nits.

I added one final nit, but I'll let you decide whether you want it do it.

As I don't know Awkward very well, I can't speak to whether this is the best way to connect to it, but things appear reasonable.  So, I'm approving what I'm able to review.

graphblas/io.py
jim22k: Good call. I updated to make the class global.

PR #282, opened 2022-09-30
Rework _repr_html_ for expressions

jim22k:
- Move bottom italic message into information note accessible via mouseover.
- Remove display of input values.
- Combine expression header into result header.
- Update tests

__Note__: I purposefully remove the `format` in the Expression header, as it gives further hint to the user that this is not a real value yet, so they shouldn't worry about what format it is being stored in.

## New look for HTML repr of Expressions
ExpressionHint

### When hovering over the `?`
ExpressionPopout
ExpressionPopout2
ExpressionPopout3


eriknw:
Oh, I'm so excited to give this a try!

I hope it wasn't too much gruntwork to update `test_formatting.py`.  That file isn't meant to be updated frequently, and I know it's kind of a pain even with the ability to copy/paste.

I'm curious about the "bizarro" test failures.  First time seeing those.

eriknw (approved):
Very nice.

Test failures are unrelated.

PR #283, opened 2022-10-01
Fix CI failures; fmin and fmax changed in numba 0.56.2

eriknw:
See: https://github.com/numba/numba/issues/8478

Also, clean up `dtypes` namespace while we're at it.

### DIFF
--- a/graphblas/binary/numpy.py
+++ b/graphblas/binary/numpy.py
@@ -56,6 +56,8 @@ _binary_names = {
     "copysign",
     "nextafter",
     "ldexp",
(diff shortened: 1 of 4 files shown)


jim22k (approved):
I don't love the need for `if version==<specific version>: apply hack`, but I'm okay with it for now as long as we get a real fix in the future once numpy is fixed.

eriknw:
Updated to check runtime behavior of numba (when necessary).  You know that moment when you think "aha, the walrus operator would make this better!"?  I didn't either until now

PR #286, opened 2022-10-04
Resolves #266 IndexUnary and Select in gb.op namespace

### LINKED ISSUE #266: Select and IndexUnary operators are in the `ops` namespace
gb.ops contains `colgt` and other things which I believe should be left out. This appears to be a mistake introduced in #233 during the import untangling. IndexUnaryOp should be calling `super()._initialize(include_in_ops=False)` in its `_initialize()` method.

jim22k:
Remove indexunary and select ops from gb.op namespace. This was an oversight during circular import cleanup.

### DIFF
--- a/graphblas/operator.py
+++ b/graphblas/operator.py
@@ -1466,7 +1466,7 @@ class IndexUnaryOp(OpBase):
     def _initialize(cls):
         if cls._initialized:
             return
-        super()._initialize()
+        super()._initialize(include_in_ops=False)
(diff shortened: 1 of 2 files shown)


eriknw:
LGTM. I reran a test, and now tests are passing.

|  Coverage decreased (-6.0e-05%)

Wow. Small number.  I guess there are about `1 / 6e-5 = 16667` lines.

PR #287, opened 2022-10-11
auto-walrus

eriknw:
bwahahaha!

We don't need to merge this, but I think it's interesting.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,6 +27,10 @@ repos:
     hooks:
       - id: pyupgrade
         args: [--py38-plus]
+  - repo: https://github.com/MarcoGorelli/auto-walrus
+    rev: v0.1.7
+    hooks:
+    - id: auto-walrus
   - repo: https://github.com/psf/black
     rev: 22.6.0
     hooks:
--- a/graphblas/__init__.py
+++ b/graphblas/__init__.py
@@ -131,8 +131,7 @@ def _init(backend_arg, blocking, automatic=False):
         # Already initialized with these parameters; nothing more to do
         return
 
-    backend = backend_arg
-    if backend == "suitesparse":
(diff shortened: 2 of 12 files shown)


MarcoGorelli:
Hey, thanks for trying this out! Sorry for commenting, just wanted to say that I think

> It is not used further in the function, so hiding it in a walrus is not bad. I could actually get behind this change:

is a great suggestion, thanks! I think I'll make it the default to only rewrite if the variable is only used within the if statement's body (and nowhere else within the function)

eriknw:
Yeah, I like your suggestion @jim22k.  Thanks for the quick turnaround @MarcoGorelli!

This PR has been updated to use v0.2.1 (and undid previous auto-walrus behavior).

I actually think it's reasonable.  I'm also pleased to not need to write walrus manually (for now), so this is a good way to get more comfortable with it.

eriknw:
This is in, thanks (lol) @MarcoGorelli!

I actually think this is a reasonable tool to help people learn how and when to use the walrus operator effectively.

I'll still probably write w/o the walrus operator most of the time, such as
```python
x = func()
if expr(x):
    do(x)
```
and rely on auto-walrus, b/c I think it generally reads better.  For example, I just used the walrus operator here, and you can compre with and without walrus: https://github.com/python-graphblas/python-graphblas/pull/288#discussion_r996594800

eriknw:
Heh, right, I'm surprised too 😛 

In the example I shared, I probably wouldn't have used the walrus operator if it weren't for the gentle nudge from auto-walrus to make me more comfortable with it.  Hence, I think it's fine--good, even--for auto-walrus to be fairly conservative and limited in its application.

Let's face it: writing your first walrus operator will be tricky for most people. Reading is easier by far, so auto-walrus can help get people over the hump.

MarcoGorelli:
😆 true, the first time I used I felt really young and trendy 🆒

Issue #206, opened 2022-04-11
Add GrB import and export functions


eriknw:
These were added in version 2.0 of the spec.

Unlike GxB import and export, the GrB version always copies data.  The function signatures should probably be similar to e.g. `Matrix.ss.import_csr`, and the docstrings for the GrB and GxB versions should reference each other.

eriknw:
GrB import and export currently support CSR, CSC, and COO.  I'm wondering how we could spell this:
```python
A = Matrix.from_csr(indptr, col_indices, values, ncols=ncols)
...
indptr, col_indices, values = A.to_csr()
```
or more like `A.ss.import_csr` where everything is a keyword, and export returns a dict:
```python
A = Matrix.import_csr(indptr=indptr, col_indices=col_indices, values=values, ncols=ncols)
...
info = A.export_csr()
# or
info = A.export("csr")
```
I may actually prefer the first option:
```python
def from_csr(cls, indptr, col_indices, values, *, ncols=None, name=None):
    # Determine nrows via `indptr`
    # ncols could be inferred by max value in `col_indices`

def from_csc(cls, indptr, row_indices, values, *, nrows=None, name=None):
    # Determine ncols via `indptr`
    # ncols could be inferred by max value in `row_indices`

def to_csr(self):
    ...
    return indptr, col_indices, values
```
The simpler Python APIs reflect the much simpler C APIs.  SuiteSparse's GxB import and export have many more options.

I'm also wondering: how is import/export COO different from to/from values/tuples.  Having both `from_values/to_values` and `from_coo/to_coo` seems undesirable, so maybe to/from coo can be combined with to/from values if/where it makes sense.

PR #288, opened 2022-10-17
Add `from_csr` and `to_csc`

eriknw:
Closes https://github.com/python-graphblas/python-graphblas/issues/206

TODO:
- [ ] Add docstrings.

Should we include `nrows=None` in `from_csr` and compare the value to the size of `indptr` if given?

COO_FORMAT doesn't seem to add any functionality compared to `from_values` and `to_values`, but we should check how it handles duplicates and maybe add it to `from_values`.


graphblas/matrix.py
```diff
@@ -664,6 +675,120 @@ def from_values(
...
+        # Ax = Ax.array
+        # Ap_len = Ap_len.gb_obj[0]
+        # Ai_len = Ai_len.gb_obj[0]
+        # Ax_len = Ax_len.gb_obj[0]
+        # if Ap.size > Ap_len:  # pragma: no cover
+        #     Ap = Ap[:Ap_len]
+        # if Ai.size > Ai_len:  # pragma: no cover
+        #     Ai = Ai[:Ai_len]
+        # if Ax.size > Ax_len:  # pragma: no cover
+        #     Ax = Ax[:Ax_len]
```
eriknw, line 775: @jim22k since I've been flirting with walrus recently (https://github.com/python-graphblas/python-graphblas/pull/287), do you like with or w/o walrus?  So much shorter with walrus.

jim22k: I can get behind the walrus in this case. It does seem to make it more readable. I do like your comment about the walrus. That is probably something we should always do for walrus uses in the codebase. Just to tip off readers.

jim22k (approved):

graphblas/matrix.py
eriknw: Agreed and done. Walrus can be dense, so comments as in this case are nice.

PR #289, opened 2022-10-18
Move files to `gb.core` for better tab-completion

eriknw:
@jim22k this is mostly as we discussed, but I didn't bother with a `_backends` folder yet.

I think this makes `python-graphblas` feel cleaner as a user.  (As a developer, I think it's going to take some time for me to get used to this--too much muscle memory for what is where!)


eriknw:
@jim22k what do you think about moving `lib` and `ffi` to `gb.core`?  I think they should go there.  They wouldn't be hiding, so they're still available to use or investigate for advanced users.

eriknw:
Removed 15 top-level names and added 1.  There are now 22 top-level names:
- `Matrix`
- `Recorder`
- `Scalar`
- `Vector`
- `agg`
- `backend`
- `binary`
- `config`
- `core`
- `dtypes`
- `exceptions`
- `indexunary`
- `init`
- `io`
- `monoid`
- `op`
- `replace`
- `select`
- `semiring`
- `ss`
- `unary`
- `viz`

jim22k:
> @jim22k what do you think about moving `lib` and `ffi` to `gb.core`? I think they should go there. They wouldn't be hiding, so they're still available to use or investigate for advanced users.

That sounds fine to me.

jim22k (approved):

PR #300, opened 2022-10-22
Allow easier array UDTs, such as `"INT64[3, 3]"`

eriknw:
This matches the "nicer" repr we have for array UDTs that we already have, so it should be natural to use.

I'm always a little skeptical when extending what numpy does, but I think this is so much easier and more readable.

Quick: can you create a valid numpy array dtype from memory?

### DIFF
--- a/graphblas/dtypes.py
+++ b/graphblas/dtypes.py
@@ -91,6 +91,12 @@ def register_anonymous(dtype, name=None):
         if isinstance(dtype, dict):
             # Allow dtypes such as `{'x': int, 'y': float}` for convenience
             dtype = _np.dtype([(key, lookup_dtype(val).np_type) for key, val in dtype.items()])
+        elif isinstance(dtype, str) and "[" in dtype and dtype.endswith("]"):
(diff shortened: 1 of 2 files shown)


eriknw:
It should be noted that record dtypes in numpy can be indexed.  For example:
```python
>>> udt = gb.dtypes.lookup_dtype({"x": int, "y": float})
>>> udt
{'x': INT64, 'y': FP64}
>>> udt.np_type
dtype([('x', '<i8'), ('y', '<f8')])
>>> udt.np_type["x"]
dtype('int64')
>>> udt.np_type[0]
dtype('float64')
```
Array dtypes cannot be indexed:
```python
>>> udt = gb.dtypes.lookup_dtype("INT64[3]")
>>> udt
INT64[3]
>>> udt.np_type
dtype(('<i8', (3,)))
>>> udt.np_type[0]
KeyError: "There are no fields in dtype ('<i8', (3,))."
```
Might we want to index our dtypes someday, and, if so, does this change our mind about approving this PR?

Issue #172, opened 2022-03-10
ENH: Add "pure GrB" backend that doesn't rely on SuiteSparse-specific behavior


eriknw:
It would be good discipline for `grblas` to have a backend option that doesn't rely on SuiteSparse-specific functions and behaviors.  This will make it easier for `grblas` to support other GraphBLAS libraries in the future.

I believe we've done a pretty decent job identifying and isolating SuiteSparse-specific functionality.  Much is already in `.ss` attributes.  Others are identified by `# SS, SuiteSparse-specific` comments.

To support such a backend, we can still use SuiteSparse:GraphBLAS, but create a `lib` object that doesn't have `GxB` items.

PR #307, opened 2022-10-27
Begin adding `suitesparse-vanilla` backend

eriknw:
This begins to address #172 by adding a `"suitesparse-vanilla"` backend that does not have `GxB` functions.

This is a WIP.  Most of these changes are straightforward. My main question is the recipe for subassign.

When this is ready to merge, I will want to move suitesparse-specific tests to their own files.


eriknw:
@jim22k, much to my surprise, this is ready to review.  I know we have a lot going on right now, so it would be nice to try to get this in within a month (or two).

I would appreciate a thorough review and pointers for when code needs better comments. Many changes are mundane, but some aren't. I plan to give this another once-over too and look at the coverage.

This doesn't get us to 100% vanilla. For example, we still have GxB operators, and some tests (like formatting) don't run with the suitesparse-vanilla backend. But, this PR is still worth merging. It gets us a long way to supporting vanilla GraphBLAS. It pays off a lot of debt, and it makes it harder to incur more debt b/c we test vanilla. I also fixed a bug or two.

The subassign recipe should be verified. It passes tests, but I don't completely trust it.

I think someday we should abstract away how we use `ffi` so we could, in principle, support a backend that doesn't use cffi.

eriknw:
I added tests to stress the vanilla recipe for subassign, and they're passing.

All that is left before merging is a final skim through the changes and a peek at coverage.

eriknw:
@jim22k I think this is ready to go in.

This includes a fair number of formatting changes (which is natural to do when going over the whole codebase), which makes reviewing more cumbersome (sorry!).

This fixes a couple bugs, adds a couple enhancements, and adds a little more coverage.

graphblas/tests/test_matrix.py
```diff
@@ -950,6 +956,98 @@ def test_assign_row_col_matrix_mask():
         C[0, 0](B.S) << 100
 
 
+@pytest.mark.parametrize("index", [slice(12), list(range(12))])
+def test_subassign_combos(index):
+    #         0  1  2  3  4  5  6  7  8  9 10 11 12 13
+    # mask    1  1  1  1  0  0  0  0  _  _  _  _
+    # val     1  2  _  _  3  4  _  _  5  6  _  _
+    # self   10  _ 20  _ 30  _ 40  _ 50  _ 60  _ 70  _
```
eriknw, line 964: btw, this is the test that gives us confidence in the vanilla subassign recipes. The recipes pass tests, so who needs to actually grok them!

jim22k (approved):

PR #310, opened 2022-10-28
Add e.g. `from_dcsr` and `to_dcsc`

### LINKED ISSUE #297: Add `Matrix.from_hypercsr`, etc.
If the backend is suitesparse, these should use the `ss` methods.  For vanilla (which we can leave as TODO), I think we should add recipes that can handle hypersparse.

What names to use--`from_hypercsr` or `from_dcsr`?

eriknw:
Closes #297 

This will also make it easier to support vanilla GraphBLAS (#307). For example, converting between awkward array can use this API.

### DIFF
--- a/graphblas/core/automethods.py
+++ b/graphblas/core/automethods.py
@@ -258,6 +258,14 @@ def to_csr(self):
     return self._get_value("to_csr")
 
 
+def to_dcsc(self):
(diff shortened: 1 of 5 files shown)


jim22k (approved):

eriknw:
Thanks for the lightning-fast review!

PR #313, opened 2022-10-30
Add `from_dict` and `to_dict` methods

eriknw:
@jim22k I want your approval, because this adds methods to `Vector` and `Matrix`.

I forget why I did this, but it seems like a natural connection to make to convert to/from Python objects.

Most of the recipes are actually non-trivial to make fast. Even though this adds to our API and we want to try to avoid API bloat, I think these are worthwhile, b/c they are useful, intuitive, fast, tested, document, and difficult for a user to get right (or as good) on their own.

(Comment: boy am I glad I created `scripts/autogenerate.py`--this has paid dividends!)


jim22k (approved):
Since NetworkX allows a dict-of-dict in the primary Graph constructor, I'm okay including it in python-graphblas.

PR #321, opened 2022-11-12
Update to SuiteSparse:GraphBLAS 7.3.2

eriknw:
This adds `gb.ss.about["openmp"]`, which indicates whether SuiteSparse:GraphBLAS was compiled with openmp enabled.

### DIFF
--- a/graphblas/ss/_core.py
+++ b/graphblas/ss/_core.py
@@ -203,6 +203,9 @@ class About(Mapping):
         "api_url": lib.GxB_API_URL,
         "compiler_name": lib.GxB_COMPILER_NAME,
     }
+    _bool_options = {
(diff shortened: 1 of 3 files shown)


eriknw:
You know, this raises the question of what our support policy is for SuiteSparse:GraphBLAS.

Is it okay to support only the latest version of SuiteSparse:GraphBLAS if its API changes? Should we try to take care to support a range (such as the last 6-18 months) of SuiteSparse:GraphBLAS?

For now, I think it's best to support what is easiest, which may be restricted to the very latest version. In the future, we may want to be more accommodating.  For example, a system admin may install SuiteSparse:GraphBLAS for all users and pin to a specific version for a period of time, but users may be able to update their Python environments more frequently.

PR #322, opened 2022-11-13
Support using `min` and `max` builtins as operators

eriknw:
This also adds support for a few other builtin operators such as `any` and `all`. I want `min` and `max`, because it will make for faster/smoother demos to show people.

@jim22k, please let me know if you don't want to add support for any of the other builtins.  I really want `min` and `max` though, and supporting the others makes sense to me.


jim22k:
I think for now we should only support `min`/`max` and maybe `abs`. Those are uncontroversial.

`pow` is somewhat obscure, so it might be okay to include, but not urgent.
`sum` and `all` are spelled differently than the GraphBLAS `plus` and `land`, so we should think about aliasing. However, as you mentioned, aliasing happens at different levels -- just the BinaryOp or also the Semiring? What about string finding (i.e. "sum.land" as a semiring)?

`any` is the most problematic because it has a different meaning that Tim's GxB extension for semiring "any_plus" which is really a "pick any", not the Boolean meaning of "logical_or" that Python uses. I'm not sure how to solve that without renaming the GxB stuff.

PR #329, opened 2022-11-14
Try using mamba instead of conda in CI

eriknw:
🤞


eriknw:
I think this is a little bit faster.

eriknw:
Btw, I think I disabled `codecov` message in the "vanilla" PR.

Issue #290, opened 2022-10-19
Expose `GxB_Vector_sort` and `GxB_Matrix_sort`


eriknw:
This is similar to `.ss.compactify`, which we found useful.  The GxB_sort methods accept a binary operator, and can simultaneously return the index permutations and the sorted values.

PR #333, opened 2022-11-18
Add GxB sort methods

eriknw:
Closes https://github.com/python-graphblas/python-graphblas/issues/290. I've been sitting on this for a couple weeks. It still needs tests and documentation, but everything seems to work.

@jim22k, would you like to take a moment to consider the options of having `foo_rowwise` and `foo_columnwise` methods vs having `foo(order="rowwise")`? Here are the methods for each convention, including `sort_rowwise` as in this PR.

These have `order="rowwise"` keyword argument:
- `Matrix.from_dicts`
- `A.to_dicts`
- `A.ss.flatten`
- `A.ss.reshape`
- `v.ss.reshape`

These use `_rowwise` and `_columnwise` suffix:
- `A.reduce_rowwise`
- `A.ss.scan_rowwise`
- `A.ss.selectk_rowwise`
- `A.ss.compactify_rowwise`
- `A.ss.sort_rowwise`

Why not `A.ss.flatten_rowwise` and `A.ss.flatten_columnwise`?
What do we think about e.g. `A.reduce(order="rowwise")`?

Also, this adds some functionality with descriptors. Cleaning up these descriptors can provide a path for implementing https://github.com/python-graphblas/python-graphblas/issues/20.

Finally, what names should we give the objects returned from `sort`? It's unusual to have an API that returns multiple Matrix or Vector objects.

### DIFF
--- a/graphblas/core/base.py
+++ b/graphblas/core/base.py
@@ -4,14 +4,13 @@ from .. import backend, config
 from .. import replace as replace_singleton
 from ..dtypes import BOOL
 from ..exceptions import check_status
-from . import NULL, ffi
+from . import NULL
 from .descriptor import lookup as descriptor_lookup
 from .expr import AmbiguousAssignOrExtract, Updater
 from .mask import Mask
 from .operator import UNKNOWN_OPCLASS, binary_from_string, find_opclass, get_typed_op
 from .utils import _Pointer, libget, output_type
 
-CData = ffi.CData
 _recorder = ContextVar("recorder")
 _prev_recorder = None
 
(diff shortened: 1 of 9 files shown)


eriknw:
I think this is good to go.  @jim22k, any preference for order or return arguments? permutations then values, or values then permutations?

jim22k:
I think values, then permutations for return order.

eriknw:
> I think values, then permutations for return order.

Done. Thanks @jim22k. SuiteSparse:GraphBLAS also uses values then permutations as the return order, so we should probably follow what Tim does unless we have a good reason otherwise.

I did change one API behavior. We will return `(None, None)` if both `values=False` and `permutations=False`, but SuiteSparse:GraphBLAS would raise in this case.

jim22k (approved):

PR #338, opened 2022-11-21
Update docs/env.yml

eriknw:
Hopefully the environment will be solvable fast enough

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,11 +5,12 @@ version: 2
 sphinx:
   configuration: docs/conf.py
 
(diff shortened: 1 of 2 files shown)


eriknw:
It was taking too long to create the environment on readthedocs. This PR makes three changes to improve things:
1. Use `mamba` instead of `conda`
2. Add `nodefaults` channel (i.e., only use packages from conda-forge)
3. Install from `conda-forge` instead of pip where possible and up to date

Adding `nodefaults` actually makes solving the environment pretty fast for both conda and mamba. I suspect mamba downloads files faster.

PR #339, opened 2022-11-22
Update "Development Status" classifier to "Production/Stable"

eriknw:
I like this explanation: https://martin-thoma.com/software-development-stages/#5-productionstable

I think `python-graphblas` is stable enough to reliably use and ready for production.

@jim22k do you concur?


jim22k:
Agreed

PR #343, opened 2022-12-06
Add implementation-specific descriptors

### LINKED ISSUE #20: Add implementation-specific descriptors

eriknw:
Closes #20.

@jim22k, I have a few questions:
1.  `axb_method=` or `AxB_method=`?
2. What would better names be for descriptors (to show in the recorder)?
3. Should we bother with better repr for descriptors, and what would that look like?
4. I documented SuiteSparse descriptor options in `graphblas/core/ss/descriptor.py:get_descriptor`, but didn't add it to the docs website. I plan to make a new issue for documenting backend-specific descriptors. Sound good?
5. There are a couple spots where descriptors are ignored b/c they can't be used, because a backend could always have an extended API that does. Should we try to create the descriptor anyway? If we do and it's non-NULL, should we raise (b/c options aren't used)?
6. Is there anything else I can do to help make reviewing easier? I doubt this needs careful review for every change, but a second pair of eyes for "reasonableness" is always appreciated.

I think this PR covers pretty much everywhere custom descriptors may be used.

Also, I got rid of caching of custom descriptors. We previously did this b/c descriptors weren't cleaned up properly, not for speed (probably). Now they are cleaned up properly.


graphblas/core/ss/descriptor.py
```diff
@@ -13,40 +16,172 @@
...
+            False: False,
+        },
+        # GxB
+        "axb_method": {
+            lib.GxB_AxB_GUSTAVSON: "gustavson",
+            lib.GxB_AxB_DOT: "dot",
+            lib.GxB_AxB_HASH: "hash",
+            lib.GxB_AxB_SAXPY: "saxpy",
+            lib.GxB_DEFAULT: "default",
+        },
```
jim22k, line 67: How does this mapping work? It feels backwards to me.

> if key in self._enumerations and isinstance(val, str):
            val = val.lower()
            val = self._enumerations[key][val]

Shouldn't the string (ex. "gustavson") be the key and `lib.GxB_AxB_GUSTAVSON` be the value?

eriknw: In `BaseConfig.__init__`, we add the reverse mappings to `self._enumerations`, so you can use either `"gustavson"` or `lib.GxB_AxB_GUSTAVSON`. I'm not sure why `_enumerations` and `_bitwise` map the lib object to string, which feels backwards to me. I guess a possible advantage of doing so is ensuring a 1-to-1 mapping of lib to string, but I don't know why this would be important. Also, having both directions of mappings in the same dict feels weird too.

So, I'm happy to clean this up.

eriknw: Updated. There are now forward and reverse mappings.

jim22k (approved):

eriknw:
> 1. axb_method= or AxB_method=?

I decided to normalize everything to lowercase and raise if this causes duplicates. Since descriptor options are keyword arguments, they can potentially overlap with other argument names, and a workaround is to use a capitalized version of the keyword for the descriptor.

> 5. There are a couple spots where descriptors are ignored b/c they can't be used, because a backend could always have an extended API that does. Should we try to create the descriptor anyway? If we do and it's non-NULL, should we raise (b/c options aren't used)?

I decided to create the descriptor, then ignore it. This at least ensures that the descriptor options were valid, which will help us transition to when descriptors can be used.

@jim22k, auto-compute of expressions leads to an interesting edge case when it comes to custom descriptors. Suppose an expression is used causing it to auto-compute (and the result is stored as e.g. `expr._value`). Now suppose you call` .new(**opts)` on the expression giving it custom descriptors. Currently, we ignore the custom descriptors and return the already-computed `expr._value`. I think this is okay. If a workflow with a backend requires fine control with custom descriptors, then auto-compute should probably be disabled to avoid this.

I'm going to merge soon--gonna do one final scan and check coverage.

eriknw:
This is in!

This is a significant addition, and it feels great to close an issue that is 2.5 years old 😃

Issue #315, opened 2022-11-03
Add conversions to/from `pydata/sparse` to `graphblas.io`


eriknw:
Probably as `gb.io.from_pydata_sparse` and `gb.io.to_pydata_sparse`. These should handle Vector and Matrix, and raise for higher dimensions. See: https://sparse.pydata.org/en/stable/

PR #347, opened 2022-12-11
Add conversions to/from pydata/sparse to graphblas.io

SultanOrazbayev:
Resolves https://github.com/python-graphblas/python-graphblas/issues/315


jim22k:
Comments from the meeting:

- Let's avoid the numpy usage for Vectors. Creating something dense could potentially use a lot of memory.
- If already in GCXS format, call scipy.sparse conversion method (like you already have)
- If in DOK, convert to COO and then call `from_coo` (no need to involve scipy.sparse)

SultanOrazbayev:
Hmm, one of the tests failed due to a timeout during download of packages.

eriknw:
Interesting `mamba` failure:
```
error    libmamba Error when extracting package: std::bad_alloc
��a�2V ΄�2V-pyhd8ed1ab_0.co0
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
 extraction failed
/home/runner/work/_temp/c260dea5-3f7a-4222-b339-9158ddcbecbf.sh: line 51:  2132 Aborted                 (core dumped) mamba install pytest coverage coveralls=3.3.1 pytest-randomly cffi donfig pyyaml${yamlver} pandas${pdver} scipy${spver} numpy${npver} awkward${akver} networkx${nxver} numba${numbaver} "graphblas>=7.4.0"
Error: Process completed with exit code 134.
```
This is with `mamba install pytest coverage coveralls=3.3.1 pytest-randomly cffi donfig pyyaml=6.0 pandas scipy numpy awkward=1.9 networkx numba graphblas>=7.4.0` on Python 3.8.

I wonder if we should switch back to `conda`, which has gotten faster/better since we switched to `mamba`.

jim22k (approved):

eriknw:
@SultanOrazbayev, I think this is close. I added a commit to install `sparse` in CI, and tests are failing.

It looks like `sparse` objects don't have `.format` attribute. Maybe use `isinstance` instead?

SultanOrazbayev:
Yes, this is odd. Let me run a few checks locally.

eriknw:
btw, the failure was with `sparse=0.12`. We can only support `0.13`.

eriknw:
Also, how do you like the "fail fast" strategy compare to running all the tests all the time? My thinking is that failing fast can lead to faster iterations, b/c it helps queueing lots of CI jobs.

SultanOrazbayev:
I'm not sure if I understand correctly, but it looks like only Windows-specific tests are failing. Do you know what's a good way to debug this? (I don't have access to a windows machine)

.github/workflows/test_and_build.yml
```diff
@@ -142,6 +142,7 @@ jobs:
           # Consider removing old versions when they become problematic or very old (>=2 years).
           nxver=$(python -c 'import random ; print(random.choice(["=2.7", "=2.8", "=3.0", ""]))')
           yamlver=$(python -c 'import random ; print(random.choice(["=5.4", "=6.0", ""]))')
+          sparsever=$(python -c 'import random ; print(random.choice(["=0.12", "=0.13", ""]))')
```
eriknw, line 145: This is where we can remove `"=0.12"`. I'm afraid to touch the PR while you're active ;)

eriknw, line 145: Accept this suggestion if you haven't already changed it locally :)
```suggestion
          sparsever=$(python -c 'import random ; print(random.choice(["=0.13", ""]))')
```

SultanOrazbayev:
Yes, fail fast works well for me.

eriknw:
> Yes, fail fast works well for me.

Great! Yeah, it was interesting when we were both working on PRs at the same time a couple days ago.

eriknw:
Oh, looks like it is a Windows-specific failure. That's surprising to me. Should we skip it on Windows?

SultanOrazbayev:
> Should we skip it on Windows?

Hmm, not sure.

> where <class 'graphblas.core.matrix.Matrix'> = gb.Matrix

This looks fixable (though I don't yet know how).

eriknw:
Oh, it looks like the dtypes are different. One is INT32, and another is INT64. This happens on Windows. maybe specify dtype, or just do `check_dtype=False`.

graphblas/io.py
```diff
@@ -220,6 +220,54 @@ def from_awkward(A, *, name=None):
...
+
+    if A.ndim == 1:
+        # the .asformat('coo') makes it easier to convert dok/gcxs using a single approach
+        _A = A.asformat("coo")
+        return _Vector.from_coo(
+            _A.coords, _A.data, dtype=_A.dtype, size=_A.shape[0], dup_op=dup_op, name=name
+        )
+
+    # handle two-dimensional arrays
+    import sparse as _sp
```
eriknw, line 253: Is `sp` a standard short name for `sparse`? I think I sometimes see `sp` as short for `scipy`. Also, no need for a preceding underscore within the function.

eriknw:
I hope you don't mind that I pushed a few changes:
- changed `import sparse as sp` to `import sparse`
- added tests for coverage
- change argument name of `from_pydata_sparse(s)` to `s`
- also test version 0.12 in CI
- add `sparse` to our "io" requirements

eriknw:
Oh, we should probably add PyData Sparse section to `docs/user_guide/io.rst` for this.

https://python-graphblas.readthedocs.io/en/latest/api_reference/#input-output

graphblas/io.py
```diff
@@ -482,6 +536,40 @@ def indices(self):
     return ret
 
 
+def to_pydata_sparse(A, format="coo"):
```
eriknw, line 539: Should we add `fill_value=None` to pass to the `sparse` array constructor?

graphblas/io.py
```diff
@@ -482,6 +536,40 @@ def indices(self):
...
+        from sparse import COO
+    except ImportError:  # pragma: no cover (import)
+        raise ImportError("sparse is required to export to pydata sparse") from None
+
+    format = format.lower()
+    if format not in {"coo", "dok", "gcxs"}:
+        raise ValueError(f"Invalid format: {format}")
+
+    # obtain an intermediate conversion via hardcoded 'coo' intermediate object
+    B = to_scipy_sparse(A, format="coo")
```
eriknw, line 564: If input is a `Matrix` and `format` is `"gcsx"`, maybe we should export to scipy `csr` array, which will be more efficient.

docs/user_guide/io.rst
```diff
@@ -68,6 +68,18 @@ that drop zero-weighted edges. The conversion from python-graphblas to scipy.spa
...
+
+A python-graphblas Matrix can be created from a 2-D (PyData) sparse array or matrix using
+``gb.io.from_pydata_sparse()``.
+
+``gb.io.to_pydata_sparse()`` will output a 2-D (PyData) sparse array given a python-graphblas Matrix.
+The sparse format can be specified. It defaults to "coo".
+
+*Note about zero-weighted edges in (PyData) sparse:* sparse treats missing values and zeros identically,
+the user should be aware of the potential for errors occurring when zero-weighted edges are handled by
+(PyData) sparse.
```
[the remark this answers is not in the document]
SultanOrazbayev: You are correct, the relevant keyword is `prune` (as noted during last meeting), the easiest solution is to remove the paragraph.

graphblas/io.py
SultanOrazbayev: Not sure about this one.  Because of the intermediate `scipy` conversion (which doesn't accept `fill_value`), I don't know how this would look like.

setup.py
```diff
@@ -2,8 +2,8 @@
 
 extras_require = {
     "repr": ["pandas >=1.2"],
-    "io": ["networkx >=2.8", "scipy >=1.8", "awkward >=1.9"],
-    "viz": ["matplotlib"],
+    "io": ["networkx >=2.8", "scipy >=1.8", "awkward >=1.9", "sparse >=0.12"],
+    "viz": ["matplotlib"],  # Min required version?
```
SultanOrazbayev, line 6: Could we set this to `3.6` for now? And then work backward if it causes problems?

eriknw: Sure, but let's do `matplotlib >=3.5`. Also, this metadata has been moved to `pyproject.toml`, so it'll need to be added there to fix the merge conflict (sorry about that!).

graphblas/io.py
eriknw: I think `fill_value` would be passed to `s.asformat(format, fill_value=fill_value)` at the very end.

SultanOrazbayev: Hmm, `pydata.sparse` doesn't have that `kwarg`? https://sparse.pydata.org/en/stable/generated/sparse.SparseArray.asformat.html

eriknw: Oh, my bad, you're right. I meant e.g.:
```python
s = COO.from_scipy_sparse(B, fill_value=fill_value)
```
I think all `sparse` constructors accept `fill_value` (`.asformat` methods being the exception). Us doing this will reinforce the data model used by `sparse`.

SultanOrazbayev: Nope, that's not an option either...  https://sparse.pydata.org/en/stable/generated/sparse.COO.from_scipy_sparse.html
The converter doesn't accept any arguments other than a scipy sparse matrix.

That's also why I had the deleted paragraph about missing/zero treatment being similar to `scipy.sparse` as in this case it would inherit whatever treatment `scipy.sparse` applied (though you are correct that's not in general the case, so that paragraph was deleted).

eriknw: heh, double oops. Right, it doesn't look like `sparse` handles `fill_value` at all with `scipy.sparse` objects. Okay, then no need to add it.

eriknw (approved):

eriknw:
This looks great! Very well done.

We just need to merge main to this PR to handle the merge conflict, then we can merge.

setup.py
SultanOrazbayev: OK, I will discard these changes and re-intro them to pyproject.tom after syncing to the main.

graphblas/io.py
```diff
@@ -220,6 +220,60 @@ def from_awkward(A, *, name=None):
...
+        import sparse
+    except ImportError:  # pragma: no cover (import)
+        raise ImportError("sparse is required to import from pydata sparse") from None
+    if not isinstance(s, sparse.SparseArray):
+        raise TypeError(
+            "from_pydata_sparse only accepts objects from the `sparse` library; "
+            "see https://sparse.pydata.org"
+        )
+    if s.ndim > 2:
+        raise _GraphblasException("m.ndim must be <= 2")
```
eriknw, line 252: Btw, this isn't covered. Can we add a simple test for this (empty `sparse` object is fine)? See:

https://app.codecov.io/gh/python-graphblas/python-graphblas/pull/347

Also, I'm not sure I like `GraphblasException` here and elsewhere (i.e., `from_numpy` has the same check and exception), but we don't need to change it in this PR.

eriknw: You can also see code coverage here: https://coveralls.io/builds/56178659

The first comment in the PR by coveralls automatically updates to point to the latest coverage for the PR. I find it handy when writing tests.

SultanOrazbayev:
<details>

______________________ test_matrix_to_from_pydata_sparse _______________________

    @pytest.mark.skipif("not sparse")
    def test_matrix_to_from_pydata_sparse():
        coords = np.array([[0, 1, 2, 3, 4], [0, 1, 2, 3, 4]], dtype="int64")
        data = np.array([10, 20, 30, 40, 50], dtype="int64")
        s = sparse.COO(coords, data, shape=(5, 5))
        v = gb.io.from_pydata_sparse(s)
        assert v.isequal(gb.Matrix.from_coo(*coords, data, dtype=dtypes.INT64), check_dtype=False)
    
        t = gb.io.to_pydata_sparse(v)
>       assert t == s

graphblas/tests/test_io.py:418: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/share/miniconda3/envs/graphblas/lib/python3.8/site-packages/numpy/lib/mixins.py:21: in func
    return ufunc(self, other)
/usr/share/miniconda3/envs/graphblas/lib/python3.8/site-packages/sparse/_sparse_array.py:305: in __array_ufunc__
    result = elemwise(ufunc, *inputs, **kwargs)
/usr/share/miniconda3/envs/graphblas/lib/python3.8/site-packages/sparse/_umath.py:48: in elemwise
    return _Elemwise(func, *args, **kwargs).get_result()
/usr/share/miniconda3/envs/graphblas/lib/python3.8/site-packages/sparse/_umath.py:492: in get_result
    r = self._get_func_coords_data(mask)
/usr/share/miniconda3/envs/graphblas/lib/python3.8/site-packages/sparse/_umath.py:672: in _get_func_coords_data
    unmatched_mask = np.ones(func_array.nnz, dtype=np.bool)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

attr = 'bool'

    def __getattr__(attr):
        # Warn for expired attributes, and return a dummy function
        # that always raises an exception.
        try:
            msg = __expired_functions__[attr]
        except KeyError:
            pass
        else:
            warnings.warn(msg, DeprecationWarning, stacklevel=2)
    
            def _expired(*args, **kwds):
                raise RuntimeError(msg)
    
            return _expired
    
        # Emit warnings for deprecated attributes
        try:
            val, msg = __deprecated_attrs__[attr]
        except KeyError:
            pass
        else:
>           warnings.warn(msg, DeprecationWarning, stacklevel=2)
E           DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
E           Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

/usr/share/miniconda3/envs/graphblas/lib/python3.8/site-packages/numpy/__init__.py:297: DeprecationWarning

</details>

Yikes. What do we do here? Pushing fixes to sparse will only fix it for the later versions... More generally, how do we handle these version-specific warnings/inconsistencies during the random combination testing?

eriknw:
> Yikes. What do we do here? Pushing fixes to sparse will only fix it for the later versions... More generally, how do we handle these version-specific warnings/inconsistencies during the random combination testing?

Yay! I recently added `-Werror` to pytest commands to catch warnings such as `DeprecationWarning`. In a _very_ surprising state of affairs, there _were_ no warnings to suppress. In general, I would expect there to be warnings out of our control that we need to suppress while testing.

So, I'm very glad to have this example to demonstrate how to suppress specific warnings. I added the necessary things to `pyproject.toml` and links to documentation to explain how to use it. It's also possible to suppress warnings for specific tests using a pytest fixture.

eriknw:
This is in! :tada: 🚀 🎆

PR #348, opened 2022-12-20
Remove pygraphblas conversions

eriknw:
`pygraphblas` appears to be defunct, and we have been unable to test these conversions for a very long time. I don't think it's necessary to deprecate these first, since they are and have been unusable.

The benefit of this is cleaner code, cleaner tab-completion, and easier maintenance.

We can revisit this if `pygraphblas` is ever revived (but I hope we can redirect effort to `python-graphblas`, which I believe is much further along in development).


jim22k (approved):

eriknw:
Thanks for the quick review @jim22k! Merging, since this also fixes CI.

PR #349, opened 2022-12-20
Use `np.bool_` instead of `np.bool8`, which is deprecated

eriknw:
NumPy 1.24 deprecates `np.bool8`, which is an alias of `np.bool_`:
https://numpy.org/devdocs/release/1.24.0-notes.html#np-str0-and-similar-are-now-deprecated


eriknw:
LGTM, thanks @eriknw!

Just to be clear, we're still on NumPy 1.23.5 due to numba.

PR #350, opened 2022-12-27
Add extra (and not by default) dev dependencies to environment.yml

eriknw:
These dependencies aren't typically needed, so I don't want to always install them. But, I want a convenient way to to make environments with some or all of these dependencies, so let's add them commented out by default. I would say these are for advanced users/devs, so I don't want to bother with breaking these up into logical groups (somebody else is welcome to do so). Presumably, if you uncomment an extra dependency to install, you should know why you want it.

@jim22k @SultanOrazbayev is there anything else you would want in one of your dev environments? I think having this section lets us be more accommodating without being overly scrutinizing.

Happy to answer questions for why I added what I did.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
       - id: autoflake
         args: [--in-place]
   - repo: https://github.com/pycqa/isort
-    rev: v5.11.3
+    rev: 5.11.4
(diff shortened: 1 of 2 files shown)


SultanOrazbayev:
Thank you, @eriknw ! This looks very comprehensive, I don't have any other dependencies to add.

PR #354, opened 2022-12-29
Add config (and optional pre-commit) for running pylint

eriknw:
Also, add a github action to run pre-commit.

There's a lot that `pylint` doesn't like about our code, and a lot about `pylint` that I don't like (by default).  Nevertheless, it can sometimes have helpful tips to improve code quality.

We don't enforce it's usage, because doing so would be cumbersome, and running it is pretty slow. I know some people really like to deveop against `pylint`, so let's add a baseline config to make this sane.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,4 +16,4 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
-      - uses: pre-commit/action@v2
+      - uses: pre-commit/action@v3.0.0
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,19 @@
+name: Lint via pre-commit
+
+on:
+  pull_request:
+  push:
+    branches-ignore:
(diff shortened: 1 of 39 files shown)


eriknw:
Unlike `pylint`, `ruff` is super-duper incredibly fast! I just discovered `ruff` and added it to this PR. We can probably use `ruff` to replace a few of our other pre-commit hooks.  Let's keep an eye on `ruff`, as I don't know of many scientific packages using it yet.

I know this much tooling for hygiene may feel like overkill, but I think they will prove useful as we update to support newer and newer Python versions. For example, it won't be long until we'll be supporting Python 3.11 through 3.14, and all of these tools allow specifying the minimum Python version to support.  Soon we ought to specify `strict=True` in `zip`, and these tools will tell us when.  And, of course, they ensure new code matches the hygiene of the current code.

On the flip side, I don't want any of us to feel like we're overly controlled and encumbered by the tooling (which overly-strict `pylint` can be like), so let's see how it goes and adjust accordingly.

I think this PR improves things, so I plan to merge it soon.

PR #362, opened 2023-01-06
Remove versioneer.py; use setuptools-git-versioning instead

eriknw:
There are a lot of modern alternatives to versioneer.py that are much less invasive. I think I like `setuptools-git-versioning`:

https://setuptools-git-versioning.readthedocs.io/en/stable/differences.html

This works as expected for me (locally, wheel builds, etc.), but readthedocs wasn't happy for some reason. Does readthedocs install the package (editable or otherwise) before trying to build the docs?


eriknw:
This is good to go and is one step towards modernization.

Note that you now need to install `python-graphblas` to have `__version__` available. Editable installs work fine. If it's not installed, then `gb.__version__` will raise with an informative message.

We may want to have readthedocs use pip instead of conda. It doesn't seem to know how to install the package to a conda environment (there's a config option on the readthedocs website). It doesn't need `__version__` though, so it runs fine.

Also: wow, this removes 2428 lines!

eriknw:
Also, I chose `setuptools-git-versioning` b/c it's what bokeh [switched to](https://github.com/bokeh/bokeh/pull/12306), satisfies all our needs, and otherwise seems pretty nice.

eriknw:
`setuptools_scm` (a very similar tool with similar capabilities and limitations) has this to say about versions for readthedocs:
https://github.com/pypa/setuptools_scm/#usage-from-sphinx
```python
# contents of docs/conf.py
from importlib.metadata import version
release = version('myproject')
# for example take major/minor
version = '.'.join(release.split('.')[:2])
```
This still implies the package was installed (perhaps editable), which I'm not sure is happening.

eriknw:
I figured out the readthedocs issue. It's amazing what you can accomplish by reading the docs and trying things! Merging.

Issue #355, opened 2022-12-29
Achieve 100% code coverage


eriknw:
Let's aspire to reach 100% code coverage when running the test suite.  If we can hit this, then let's (probably) enforce it.

Coverage reports are here:
- https://coveralls.io/github/python-graphblas/python-graphblas
- https://app.codecov.io/gh/python-graphblas/python-graphblas

At times, it is appropriate to add `# pragma: no cover` or `# pragma: no branch` on uncovered code.  When this is done, we should try to provide a short annotation or description for why it's not covered. For example, we use `# pragma: no cover (numba)` for functions jitted with numba. Here is a list of annotations we currently use:
- `numba`: function is jitted with numba
- `sanity`: this code should _never_ be possible to be reached, but if it is, we want a hard failure
- `safety`: reasonably paranoid code that is difficult to test, but is desirable for defensive, maintainable code
- `suitesparse`: more specific than `safety`, this code may be reached by an error originating from SuiteSparse:GraphBLAS that is difficult for us to trigger
- `deprecated`: the code is deprecated and we don't care about its coverage
- `debug`: code helpful for debugging and is not covered in normal usage; often in tests
- `import`: code branch depends on the (un)availability of an optional package
- `ipython`: code that is exercised when used in IPython or Jupyter and is hard for us to test
- `unused`: code that could be safely removed (and maybe we should), but we're keeping it for possible future use
- `???`: the code _ought_ to be reported as covered, but `coverage` may not be working properly

We should aspire to not have bare `# pragma: no cover`. Let's try to always have a reason.

I marked this as "good first issue", b/c some coverage should be straightforward, and I would be _delighted_ to see tests added for the sole purpose of adding coverage.

I also marked this as "good second issue", b/c some coverage will probably be more difficult to achieve.

If anybody has questions for how to cover specific lines, feel free to respond here, ask us on Discord, or join our weekly community call.

We're currently above 99%, so we're not far off!

eriknw:
I don't think we should enforce a coverage limit. This is not friendly to new contributors, and I think it's more valuable to have randomized tests as currently done, which comes at the cost of occasional coverage misses.

jim22k:
Agree. High level of coverage is nice, and we should be aware of significant drops in coverage (and open future PRs to fix). But strict 100% requirement is not worth the cost in my opinion.

PR #365, opened 2023-01-07
Change CI matrix to be [os]x[slowtask]

eriknw:
Randomize Python version and from where to install `python-suitesparse-graphblas`.

The full matrix is over what is more important: that all tests (slow tests, notebooks, etc.) are run on all OSes. This should ensure that our code coverage doesn't fluctuate based on random chance (as it currently does), which will help us achieve and _enforce_ 100% code coverage (#355).

We actually don't have a strong dependence on Python version and the source of `python-suitesparse-graphblas`, so it makes more sense to randomize on those. We still want to test them though, so randomizing makes the most sense imho.

I consolidated the logic of what is installed by conda (_ahem_, mamba). This simplifies the CI config overall, but it makes for a pretty large `mamba install` statement (which we'll be able to simplify once we have wheels for Windows and OS X).

This drops us down to 12 CI jobs in the main scripts (from 15 jobs) :tada: . It would be easy to up it back to 15 if desired. Another benefit of this form is that we can add additional Python versions (such as Python 3.11, PyPy, etc.) without increasing the number of CI jobs.

### DIFF
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -40,7 +40,7 @@ jobs:
             1
             1
             1
-      - name: RNG for source type of python-suitesparse-graphblas
+      - name: RNG for source of python-suitesparse-graphblas
         uses: ddradar/choose-random-action@v2.0.2
         id: sourcetype
         with:
@@ -52,6 +52,26 @@ jobs:
           weights: |
             1
             1
+            1
+            1
+      - name: RNG for SciPy version
+        uses: ddradar/choose-random-action@v2.0.2
+        id: spver
+        with:
+          contents: |
+            '=1.7.0'
+            '>1.7.0'
+          weights: |
+            1
+            1
+      - name: RNG for Awkward version
+        uses: ddradar/choose-random-action@v2.0.2
+        id: akver
+        with:
+          contents: |
+            '<2'
+            '>=2'
+          weights: |
             1
             1
       - name: Setup conda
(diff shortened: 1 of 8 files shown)


eriknw:
CC @jim22k. I don't think this needs careful review, but I want to share b/c I think it's cool!

I knew I would like randomization in tests :) . We now randomize the following:
- Python version: `[3.8, 3.9, 3.10]` (and soon `3.11`)
- Source of `python-suitesparse-graphblas`: `[conda-forge, wheel, source, upstream]`
- Backend: `[suitesparse, suitesparse-vanilla]`
- Blocking: `[True, False]`
- Recording: `[True, False]`
- `mapnumpy`: `[True, False]`
- And random order of tests via `pytest-randomly`

Ignoring `pytest-randomly`, this gives us `3 * 4 * 2 * 2 * 2 * 2 = 192` possibilities (256 once we add Python 3.11). And yet I feel comfortable sampling only 12 at a time.

We could also randomize dependency versions, such as:
- `scipy`: `[==1.7.0, >1.7.0]`
- `awkward`: `[<2, >=2]`
- `numba`: `???`

If we have a minimum version for a dependency, I think it's good practice to test against it.

eriknw:
I'd like to go ahead and give this a try to see if it achieves its primary goal: consistent coverage.

Compared to a month ago, our CI is faster wallclock, faster total time, does more (e.g., now does pylint, pre-commit, etc), tests more combinations of things, has more reliable coverage, and has fewer CI jobs :tada:

This PR is probably a little bit overkill the way I do correlated randomness, but I think e.g. being careful about testing both backends will be useful when we flesh out the `suitesparse-vanilla` backend.

Looking ahead, since we have 12 main CI jobs, we can reasonably add up to 8 more jobs for other things such as new backends (we get 20 runners in the free tier). We can also add additional Python versions and dependency versions without increasing the number of jobs.

I like the idea of testing dependency versions, but this requires different versions for each Python version, so I felt it was too complicated to add to this PR.

One final note: I did find a couple tests whose coverage depends on the order that tests run. This isn't ideal, but I dealt with the ones I found (but they still depend on order).

PR #366, opened 2023-01-08
Move suitesparse-specific operators to `.ss` namespace (and deprecate current location).

### LINKED ISSUE #356: Use `pyproject.toml` instead of `setup.py` and `setup.cfg` where possible
Let's become more modern and update to use `pyproject.toml` instead of `setup.py` and `setup.cfg`. We'll probably still need `setup.cfg`, but let's move as much as we can to `pyproject.toml`.

We should also see if we can move other configuration such as `.readthedocs.yml` and `.codecov.yml` to `pyproject.toml`.

This relates mostly to PEP 621, and also to PEPs 518, 517, 660, and maybe others:
- https://peps.python.org/pep-0621/
- https://peps.python.org/pep-0518/
- https://peps.python.org/pep-0517/
- https://peps.python.org/pep-0660/

I labeled this as a "good first issue" _if_ you're knowledgeable about such things or want to learn about it, since it's mostly about moving metadata from one format to another. This is a useful skill that I'm sure could be applied to many other projects as well!

These PEPs are largely about tooling for building packages and making metadata less programmatic. Since `python-graphblas` is pure Python and our releases, builds, and uploads are automated, we don't need to care that much about build tooling. What we care about with this is updating and centralizing metadata (i.e., PEP 621).

More context:
- https://www.shore.co.il/blog/pep-621-and-setuptools/
- http://ivory.idyll.org/blog/2021-transition-to-pyproject.toml-example.html
- https://snarky.ca/what-the-heck-is-pyproject-toml/

### LINKED ISSUE #364: Move suitesparse-specific operators to `.ss` namespace (and deprecate current location)
SuiteSparse-specific operators such as `any_firsti` that can't be recreated using UDFs should be moved to `ss` namespace and their current location deprecated.

For example, use `gb.semiring.ss.any_firsti` instead of `gb.semiring.any_firsti` (which should warn upon use).

This will help `python-graphblas` support other GraphBLAS implementations in the future.

eriknw:
Closes #364

**Edit:** also closes #356 

I kinda wish we had done this from the very beginning to better enable support for different implementations.

So far I only moved "positional" operators such as `positioni` and `firstj`, b/c I think we can _probably_ make UDFs for the rest.

To enable deprecations, the current code is somewhat complicated. We'll be able to clean it up again once the deprecated methods are removed in eight-ish months.

This ended up required quite a few changes. I relied on having good test coverage to drive what changes need made. It's very possible I missed some corner case somewhere 🤷‍♂️ 

I think it would be nice if the GraphBLAS spec had positional operators in some way, but it's TBD whether they should be as e.g. `BinaryOp` or `IndexBinaryOp` and maybe `IndexSemiring`.


eriknw:
I would like us to explore whether we can come up with good vanilla recipes for the aggregators `first`, `last`, `first_index`, `last_index`, `argmin`, and `argmax`. That way, we wouldn't need to move them to `.ss`. They seem doable at first glance, but they may not be very efficient.

eriknw:
Before we merge or release this, I want to verify that we can create UDFs with numba to create all the operators we may need in order to support other backends. This exercise will also be useful for #298.

eriknw:
Alright, I think this is ready if anybody wants to skim through it (CC @jim22k)

jim22k (approved):

jim22k:
A surprising amount of changes needed, but I agree this will help with supporting multiple backends.

eriknw:
> A surprising amount of changes needed

Right?!

> but I agree this will help with supporting multiple backends.

Yup. It'll be particularly interesting to support a non-cffi backend. Still plenty to do to get there.

Thanks for reviewing @jim22k! Merging.

PR #368, opened 2023-01-10
Randomize version of dependencies

eriknw:
Let's see how this goes...


eriknw:
Alright, this was helpful to determine minimum versions that we require for `numpy` and `networkx`.

I prefer to have the randomness to be done via github action instead of Python (as is done for NumPy) so it's easier to determine versions in CI jobs (and also easy to set random weights to 0). NumPy versions are done by Python so that we can have different versions for different Python versions.

Also, I didn't want to directly wade into the dependency relationship between `numpy`, `scipy`, and `numba`, but randomizing versions of `scipy` and `numba` may also be nice, especially since we rely pretty heavily on numba.

I think this complication is beneficial, and I don't suspect it will be too much of a maintenance burden.

One more thing: I was getting errors when building `python-suitesparse-graphblas` from source (or upstream) when using older numpy versions. I don't know why. It would be nice to get this working, but I don't have time right now to look into it.

eriknw:
I've stressed tested this enough to trust it. It's working great.

I expect _updating_ versions of dependencies to be easier. There's no rush to remove old dependencies unless they're causing issues--they'll be removed eventually by dropping support for old Python versions. If we can, though, we should try to support major versions of dependencies for at least two years.

I expected randomizing dependency versions to be complete overkill. After doing it, I think it's very valuable and works pretty well. We now have lower limits for all our core dependencies, but we don't have min versions for `matplotlib` and `pytest` 🤷‍♂️

PR #370, opened 2023-01-11
Try switching back to conda (from mamba)

eriknw:
`conda` has gotten faster (such as parallel downloads) and it's probably what most of our users use, so let's try switching back to `conda` from `mamba`.

We first switched to `mamba` in #329 and #338.

We have occasionally encountered timeout errors when using `mamba`, and the warnings on CI are pretty annoying.


eriknw:
`mamba` is still worth using for performance reasons. Closing.

PR #373, opened 2023-01-21
More linting

eriknw:
Adopting linting standards aims to increase the maintainability, quality, and readability of the projects under `python-graphblas`. I plan to move _some_ of these rules and tools over to `graphblas-algorithms` too.

Let me know if I'm ever going overboard!


eriknw:
gravy

PR #374, opened 2023-01-22
Add `Matrix.from_edgelist` and `matrix.to_edgelist`

### LINKED ISSUE #372: Creating Matrix from an edgelist?
Do you think it's useful to have a method for creating a matrix via an edge list? This is the rough API:

```python
from graphblas import Matrix


edge_list = [(0,1), (0,2), (0,3)]
A = Matrix.from_edge_list(edge_list, edge_value=2, nrows=4, ncols=4, name='test')
# these would be all the possible kwargs
# edge_value - default is 1
# nrows/ncols - default would be imputed from the list
```

To give context why this might be useful, I'm writing a small tutorial and it seems that having an edge-list based matrix creation might be more intuitive (to avoid introducing coo/csr/dcsr/etc notations, which are not IMO beginner-friendly).

eriknw:
Closes #372.


SultanOrazbayev (commented):
LGTM!

I didn't think about returning the values (and using numpy arrays), but this implementation is much better.

eriknw:
Thanks for the quick  review @SultanOrazbayev!

I just added a little more functionality: `from_edgelist` can now also accept triples of `(row, colulmn, value)`. So, both of these are valid:
```python
Matrix.from_edgelist([[1, 2], [3, 4], [5, 6]], [10, 20, 30])
Matrix.from_edgelist([[1, 2, 10], [3, 4, 20], [5, 6, 30]])
```
What do you think? Is the function still clear enough?

SultanOrazbayev:
> I just added a little more functionality: `from_edgelist` can now also accept triples of `(row, colulmn, value)`. So, both of these are valid:
> 
> ```python
> Matrix.from_edgelist([[1, 2], [3, 4], [5, 6]], [10, 20, 30])
> Matrix.from_edgelist([[1, 2, 10], [3, 4, 20], [5, 6, 30]])
> ```
> 
> What do you think? Is the function still clear enough?

Yes, this looks great!

eriknw:
This may quickly become my favorite way to create matrices :)

`edgelist` is _probably_ the right term. It's very graph-centric, though, so is there a better name for these functions? Is there a more linear algebra-centric term?

Notably, there isn't a method to create a Vector from `(index, value)` pairs. If we were to add this, would we call it `Vector.from_edgelist`? What the heck is an edge for a vector?

SultanOrazbayev:
> This may quickly become my favorite way to create matrices :)
> 
> `edgelist` is _probably_ the right term. It's very graph-centric, though, so is there a better name for these functions? Is there a more linear algebra-centric term?
> 
> Notably, there isn't a method to create a Vector from `(index, value)` pairs. If we were to add this, would we call it `Vector.from_edgelist`? What the heck is an edge for a vector?

I think having `.from_edgelist` for a vector is ok. To motivate this, imagine that the vector is a 'slice' from the adjacency vector, so the corresponding values are the edges (it's just that we don't need to specify the constant node that corresponds to the slice). This might not make much sense in other vector use cases (outside of graphs), but AFAIU in `graphblas` that's the main usage of vectors (i.e. it's basically a matrix, where one dimension is removed as it's associated with a constant node).

SultanOrazbayev:
>  Is there a more linear algebra-centric term?

Perhaps`.from_elements` or `.from_matrix_elements`?

eriknw:
`Vector.from_pairs` added for consideration.

jim22k:
From the meeting discussion today, we agreed:

- Matrix.from_edgelist will accept:
   - Python pairs of edges (start, end), followed by a values list/scalar
   - Python triples (start, end, value)
   - Numpy array of edges (start, end as nx2 column array), followed by a values array/list/scalar
- Vector.from_pairs will accept:
   - Python pairs of (index, value)
- Matrix.to_edgelist will produce:
   - numpy array of edges (nx2 column array), followed by a values array/scalar

The edge arrays must be int/uint dtype.

eriknw:
Done.

PR #379, opened 2023-02-02
`.wait` return `self` so it can be used with method-chaining

eriknw:
I've wanted method chaining like this for `wait` a few times in the past few months.


eriknw:
This raises the question whether we should do the same for `build`, `clear`, and/or `resize`.

eriknw:
Nevermind, expressions shouldn't (and don't) have `build/clear/resize`, so these shouldn't return `self`.  `expr.wait()` doesn't actually modify the object, so having `val = expr.wait()` can be a nice shorthand for `val = expr.new().wait()`.

PR #381, opened 2023-02-03
Improve construction when inferring sub-array dtype (a.k.a. array subdtype)

eriknw:
This gives us an easy, consistent way to infer sub-array dtypes upon when constructing objects. Also, this handles more cases (I apparently missed some!). See tests for examples.


graphblas/core/matrix.py
```diff
@@ -1436,12 +1430,14 @@ def from_dicts(
...
         iter_values = itertools.chain.from_iterable(v.values() for v in dicts)
         if dtype is None:
-            values = np.array(list(iter_values))
-            dtype = lookup_dtype(values.dtype)
+            values, dtype = values_to_numpy_buffer(list(iter_values), subarray_after=1)
         else:
             # If we know the dtype, then using `np.fromiter` is much faster
             dtype = lookup_dtype(dtype)
-            values = np.fromiter(iter_values, dtype.np_type)
+            if dtype.np_type.subdtype is not None and np.__version__[:5] in {"1.21.", "1.22."}:
```
eriknw, line 1437: I think this is the first time we actually use the version of a dependency. This way is pretty low-tech, and hopefully fine. If we need to do more version-dependent behavior, we'll probably want to use a helper utility such as `packaging.version.parse`.

Also, CI testing of dependency versions FTW!

We can drop 1.21 in June, and 1.22 next January.

eriknw:
I need this for another PR, merging.

PR #382, opened 2023-02-04
Add `from_dense` and `to_dense` methods

### LINKED ISSUE #378: Add easier way to create dense vectors and matrices
It may be nice to have constructors to create `Vector` or `Matrix` from "dense" a.k.a "full" data. It should be able to handle numpy arrays, list of lists, and scalars. I think scalars are important to create iso-valued objects.

Creating full objects from numpy arrays is more difficult than it ought to be in GraphBLAS. For example, see `Matrix._from_dense`.

Relatedly, we could add a method to create a diagonal matrix from an array or scalar (for example, to create an identity matrix).

These are all possible to do today, such as:
```python
A = Matrix(a.dtype, nrows=a.shape[0], ncols=a.shape[1])
A[:, :] = a

# vs

A = Matrix.from_foo(a)
```
and
```python
A = Matrix(nrows=nrows, ncols=ncols)
A << scalar

# vs

A = Matrix.from_foo(scalar, nrows=nrows, ncols=ncols)
```
and
```python
v = Vector(a.dtype, size=a.size)
v[:] = a
A = v.diag()

# vs

A = Vector.from_foo(a).diag()

# vs

A = Vector.from_coo(np.arange(n), a, size=n).diag()

# vs

A = Matrix.from_coo(np.arange(n), np.arange(n), a, nrows=n, ncols=n)

# vs

A = Matrix.from_bar(a)
```
(Note that there are suitesparse-specific constructors such as `Matrix.ss.import_fullr`)

Possible signatures:
```python
def from_foo(values, dtype=None, *, nrows=None, ncols=None, name=None):
    ...  # create full matrix from array, list of lists, or scalar
    # what about rowwise or columnwise? Should we add `order="rowwise"` keyword?

def from_bar(values, dtype=None, *, k=0, nrows=None, ncols=None, name=None):
    ... # create matrix with data on the kth diagonal from array, list, or scalar (or Vector?)
```
Also related is `gb.io.from_numpy`, whose behavior we could reconsider.

eriknw:
Closes #378.

@jim22k, I think we discussed to/from dense _years_ ago, and I think we punted on methods like in this PR b/c it's surprisingly awkward to do in GraphBLAS. This PR shows that, indeed, this is difficult to do well, so I think they're good methods to add.

I also added `sort=True` to e.g. `to_csr`, because the C spec says column indices within a row may be unsorted, but sometimes we want them to be sorted. `to_coo` already has `sort=True`, so adding sorting more places seems natural.

Currently, `from_dense` is also used to create a fully dense Vector/Matrix from a scalar, which requires the shape to be provided. I think I like this better than a separate method such as `Matrix.from_scalar(1, nrows=3, ncols=4)`.

This also improves some handling of sub-array dtypes.

I still need to add documentation. I think the APIs and behaviors need the most careful review.


graphblas/core/matrix.py
```diff
@@ -1335,30 +1336,131 @@ def from_dcsc(
...
+        to_dicts
+        to_edgelist
+        from_dense
+        io.to_numpy
+
+        Returns
+        -------
+        np.ndarray
+        """
+        max_nvals = self._nrows * self._ncols
```
SultanOrazbayev, line 1434: Not sure if this is feasible or not, but we could add a check for very large matrix dimensions: if the dimensions of the matrix are such that their dense representation would exceed the available memory, then it might be worth warning the user and raising an error.

I remember `dask.array` had a similar warning when computing large arrays.

SultanOrazbayev: Actually it turns out it was from `numpy` rather than `dask`: https://github.com/numpy/numpy/blob/8ace612c3b1bbf0bd23f10556492908afdc97207/numpy/core/_exceptions.py#L117

scripts/check_versions.sh
```diff
@@ -3,7 +3,7 @@
 # Use, adjust, copy/paste, etc. as necessary to answer your questions.
 # This may be helpful when updating dependency versions in CI.
 # Tip: add `--json` for more information.
-conda search 'numpy[channel=conda-forge]>=1.24.1'
+conda search 'numpy[channel=conda-forge]>=1.24.2'
```
SultanOrazbayev, line 6: Interesting, didn't know about this syntax for channel. I used `-c` syntax before: `conda search -c conda-forge "numpy>=1"`.

graphblas/core/matrix.py
eriknw: Very good point. I added a warning in the docstring of `to_dense`, and I changed the implementation so it will quickly raise `OutOfMemory` when using SuiteSparse:GraphBLAS. Is this enough?

How would we choose the threshold for when to raise a warning for large output arrays? Do we just choose a number such as N GB, or base it on total memory or available memory?

graphblas/core/vector.py
```diff
@@ -795,6 +795,118 @@ def from_pairs(cls, pairs, dtype=None, *, size=None, dup_op=None, name=None):
             )
         return cls.from_coo(indices, values, dtype, size=size, dup_op=dup_op, name=name)
 
+    @classmethod
+    def from_dense(cls, values, dtype=None, *, size=None, name=None):
```
eriknw, line 799: `missing_value=None` to indicate which value to drop.

graphblas/core/vector.py
```diff
@@ -795,6 +795,118 @@ def from_pairs(cls, pairs, dtype=None, *, size=None, dup_op=None, name=None):
...
+            # TODO: GraphBLAS needs a better way to import or assign dense
+            rv = cls.from_coo(
+                np.arange(values.shape[0], dtype=np.uint64),
+                values,
+                dtype,
+                size=values.shape[0],
+                name=name,
+            )
+        if size is not None and size != rv._size:
+            rv.resize(size)
```
eriknw, line 853: raise instead; mention `.resize`.

graphblas/core/vector.py
jim22k: This will allow us to replace `io.from_numpy` with this implementation.

eriknw:
`missing_value=` is a nice addition to `from_dense`. However, `missing_value=` doesn't make any sense for building a dense object from a scalar, so I feel like this is now squeezing too much functionality into `from_dense`.

Compare:
```python
def from_dense(cls, value, missing_value=None, dtype=None, *, nrows=None, ncols=None, name=None):
    # handle scalars and arrays
```
vs
```python
    def from_dense_scalar(cls, value, nrows, ncols, dtype=None, *, name=None):
    def from_dense_array(cls, values, missing_value=None, dtype=None, *, name=None):
```
I prefer two methods. But, what should we name them?

eriknw:
> I prefer two methods. But, what should we name them?

Perhaps:
- `from_dense` for arrays
- Add `fill_value=None` to the main constructor to fill with a scalar
  - e.g., `Vector(size=5, fill_value=10)`

graphblas/core/vector.py
eriknw: N/A since we now have a new method for dense from scalars.

graphblas/core/vector.py
eriknw: Using `select.valuene` after importing from dense seems to work well.

scripts/check_versions.sh
eriknw: Yeah, using `-c conda-forge` _adds_ the conda-forge channel to search in, but will still search in default channels. The syntax above only searches in the specified channel.

Scroll to the bottom here: https://docs.conda.io/projects/conda/en/latest/commands/search.html

eriknw:
Splitting `from_dense` into two functions (one for "dense from scalar") simplified the implementation, so +1 from me.

I still need to update documentation. Do we like `from_iso_value`?

graphblas/core/matrix.py
SultanOrazbayev: If I understand correctly, warning in the doc string is sufficient, because numpy will raise its own MemoryError if there is not enough available memory.

eriknw:
Updated and ready for review.

eriknw:
We discussed this in our meeting today, and we settled on `from_scalar` instead of `from_iso_value`. We should also add a couple more notes such as this being cheap / low storage in SuiteSparse:GraphBLAS, and maybe show an alternative example for how to create a new matrix with a scalar value and a mask.

We could consider adding a `mask=` keyword to `from_scalar`, but this implies a shape, which makes things a little awkward.

eriknw:
`from_iso_value` is renamed to `from_scalar`, and I made a note about SuiteSparse:GraphBLAS being efficient.

@jim22k, what note about masks did you want to add?

I think this PR is ready to merge, but I'll wait for an approval (for a couple days).

jim22k:
The note could say that if you want to create an iso-valued Matrix or Vector with the same structure as an existing object, use `obj.apply(binary.second, right=scalar)`.

jim22k (approved):

eriknw:
Thanks @jim22k and @SultanOrazbayev for your input on this PR! Even though I'm getting credit for commits, I know y'all deserve credit for helping too, because your engagement does make things better ❤️ 

This is going in! 🚀

PR #385, opened 2023-02-12
Test new wheels for `python-suitesparse-graphblas` on all OSes

eriknw:
Testing wheels from https://github.com/GraphBLAS/python-suitesparse-graphblas/pull/66


eriknw:
Converted to draft b/c we're waiting on `python-suitesparse-graphblas` to be released to main PyPI. We'll need to update from where to install wheels in github actions.

PR #391, opened 2023-02-18
Read Matrix Market with `fast_matrix_market`

### LINKED ISSUE #384: Option to read MatrixMarket with `fast_matrix_market`
See: https://github.com/alugowski/fast_matrix_market

We already have `mmread` and `mmwrite` that use `scipy.io`, but `fast_matrix_market` ought to be faster (and better?). Let's make `fast_matrix_market` an optional dependency (it has Python wheels and is on conda-forge) and figure out the best way to use it. It has docs/example of creating GraphBLAS in C++ ([see here](https://github.com/alugowski/fast_matrix_market/blob/main/README.GraphBLAS.md)), but I think we should be able to get by with the Python bindings.

One option for our API is to add `engine="auto"` keyword to these functions and try to use the fastest installed library by default.

Thanks @alugowski for creating `fast_matrix_market` and for bringing it to my attention.

eriknw:
Closes #384.

This uses `fast_matrix_market` (which looks [pretty fast](https://github.com/alugowski/fast_matrix_market)) by default if available, but will fall back to `scipy` if necessary. `scipy` is still necessary to be installed even if using "fast_matrix_market" engine.

There were three small issues I encountered that I plan to raise upstream, but I was able to work around them.

TODO:
- [x] add Matrix Market section in our IO docs and mention `fast-matrix-market` to get better performance: https://python-graphblas.readthedocs.io/en/latest/user_guide/io.html
- [x] also update docs for conversions to/from NumPy to use [`from_dense/to_dense`](https://github.com/python-graphblas/python-graphblas/pull/382)


eriknw:
CC , @alugowski `fast-matrix-market` works pretty well as a near-drop-in replacement for `scipy.io.mmread`, and I'm really happy to improve the user experience around MM files--thanks! I had to work around a few _tiny_ differences, which wasn't that difficult, so these aren't that important:
- https://github.com/alugowski/fast_matrix_market/issues/9
- https://github.com/alugowski/fast_matrix_market/issues/10
- https://github.com/alugowski/fast_matrix_market/issues/11

@alugowski, when writing Matrix Market files, do you recommend passing scipy CSR or CSC matrices instead of COO? I'm curious whether `fmm` or SuiteSparse:GraphBLAS would handle this faster.

graphblas/io.py
```diff
@@ -550,29 +550,43 @@ def to_pydata_sparse(A, format="coo"):
     if format not in {"coo", "dok", "gcxs"}:
         raise ValueError(f"Invalid format: {format}")
 
-    if format == "gcxs":
-        B = to_scipy_sparse(A, format="csr")
+    if _output_type(A) is _Vector:
+        indices, values = A.to_coo(sort=False)
+        s = COO(indices, values, shape=A.shape)
```
eriknw, line 555: btw @SultanOrazbayev I just special-cased `Vector` in `to_pydata_sparse`, b/c previously it was creating a `(1, n)`-shaped sparse array from a Vector instead of a `(n,)`-shaped array.

SultanOrazbayev: Oops, sorry about that!

graphblas/io.py
```diff
@@ -590,7 +604,21 @@ def mmread(source, *, dup_op=None, name=None):
         from scipy.sparse import isspmatrix_coo
     except ImportError:  # pragma: no cover (import)
```
SultanOrazbayev, line 605: This is a minor point (and not sure if this is in line with 'best practices'), but perhaps this import check for `scipy.sparse` can be made conditional on `engine='scipy'`? (and same for fast matrix market)

graphblas/io.py
```diff
@@ -624,8 +666,30 @@ def mmwrite(target, matrix, *, comment="", field=None, precision=None, symmetry=
...
         raise ImportError("scipy is required to write Matrix Market files") from None
+    engine = engine.lower()
+    if engine in {"auto", "fmm", "fast_matrix_market"}:
+        try:
+            from fast_matrix_market import mmwrite  # noqa: F811
+        except ImportError:  # pragma: no cover (import)
+            if engine != "auto":
+                raise ImportError(
+                    "fast_matrix_market is required to write Matrix Market files "
+                    f'using the "{engine}" engine'
```
SultanOrazbayev, line 677: Not sure to what extent DRY principle is useful here, but perhaps there is scope for generalising this check for suitable import as some utility function (as a separate PR).

SultanOrazbayev (approved):
Looks great, @eriknw !

graphblas/io.py
eriknw: Ah, I know why I did it this way--`scipy` is needed _regardless_ of backend. I added a code comment saying as much.

But, yes, I 100% agree that we should only import libraries if they are needed.

graphblas/io.py
eriknw: I'm open to such a PR :)

You may notice that my style sometimes doesn't adhere to DRY for small, straightforward (to me) things, but I suspect there are cases where things could be improved.

SultanOrazbayev (approved):

eriknw:
Thanks for reviewing @SultanOrazbayev!

I just finished the final TODO for this PR--I added documentation to `io.rst`.

PR #392, opened 2023-02-20
Use `np.random.default_rng` instead of legacy functions

eriknw:
I don't think Numba supports these yet in compiled functions, since numba uses its own RNG.


eriknw:
Not much of substance here, just cleaning some things up. Merging.

PR #393, opened 2023-02-21
Remove expired deprecation of using scipy matrix

eriknw:
Note that we now maintain a list of active deprecations on our wiki:

https://github.com/python-graphblas/python-graphblas/wiki/Active-deprecations

### DIFF
--- a/graphblas/tests/test_numpyops.py
+++ b/graphblas/tests/test_numpyops.py
@@ -1,6 +1,7 @@
 # These tests are very slow, since they force creation of all
 # numpy unary, binary, monoid, and semiring objects.
 import itertools
+import sys
 
 import numpy as np
 import pytest
@@ -10,10 +11,13 @@ import graphblas.binary.numpy as npbinary
 import graphblas.monoid.numpy as npmonoid
 import graphblas.semiring.numpy as npsemiring
 import graphblas.unary.numpy as npunary
-from graphblas import Vector
+from graphblas import Vector, backend
 from graphblas.dtypes import _supports_complex
 
 from .conftest import compute
+
+is_win = sys.platform.startswith("win")
+suitesparse = backend == "suitesparse"
 
 
 def test_numpyops_dir():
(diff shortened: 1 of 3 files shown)


eriknw:
Oh, interesting test failures (randomized testing ftw?!). This is using `python-suitesparse-graphblas` from upstream, which supports complex on Windows.

It looks like `asin(4 + 0j)` gives the wrong results on Windows in SuiteSparse:GraphBLAS. This is a known limitation. From SuiteSparse:GraphBLAS documentation:

> 5. On Windows, the casin, casinf, casinh, and casinhf functions provided by Microsoft do not return the correct imaginary part. As a result, GxB_ASIN_FC32, GxB_ASIN_FC64 GxB_ASINH_FC32, and GxB_ASINH_FC64 do not work properly on Windows. This affects the GrB/asin, GrB/acsc, GrB/asinh, and GrB/acsch, functions in the MATLAB interface. See the MATLAB tests bypassed in gbtest76.m for details, in the GraphBLAS/GraphBLAS/test folder.

So, we'll need to update tests to ignore these on Windows.

I'm not sure what's up with the error with `ldexp`. Need to print more info.

eriknw:
Heh, so we may have more places where we use `if not _supports_complex` instead of `if is_win`. If so, we'll fine them eventually.

graphblas/tests/test_numpyops.py
```diff
@@ -61,6 +65,11 @@ def test_npunary():
         "INT64": {"reciprocal"},
         "FC64": {"ceil", "floor", "trunc"},
     }
+    if suitesparse and is_win:
+        # asin and asinh are known to be wrong in SuiteSparse:GraphBLAS
+        # due to limitation of MSVC with complex
+        blocklist["FC64"].update({"asin", "asinh"})
+        blocklist["FC32"] = {"asin", "asinh"}
```
eriknw, line 72: I don't know if we strictly need to exclude `asinh` given the input arguments we use, but `asinh` was called out in the documentation for not working.

eriknw:
I'm going to go ahead and merge this since it contains a fix, and I don't think the removed deprecations are particularly significant.

Issue #396, opened 2023-03-07
[pyos] Broken link in docs


szhorvat:
On https://python-graphblas.readthedocs.io/en/latest/user_guide/fundamentals.html, the graphblas.org link is broken, probably due to a double-`:` in `https:://graphblas.org`

<img width="449" alt="image" src="https://user-images.githubusercontent.com/1212871/223452528-fcabc80e-bea7-41a6-a2df-27ef16aedea5.png">

https://github.com/pyOpenSci/software-submission/issues/81

eriknw:
Thanks for the report @szhorvat and the fix @SultanOrazbayev! :heart:

szhorvat:
I won't open a separate issue for this, as it's not a big deal, but the doc page doesn't look great in dark mode, and the header image just went missing. I think it was there earlier today.

<img width="1027" alt="image" src="https://user-images.githubusercontent.com/1212871/223567191-518312b0-1450-42f2-8815-7338f1405bfc.png">

eriknw:
> I won't open a separate issue for this, as it's not a big deal, but the doc page doesn't look great in dark mode, and the header image just went missing. I think it was there earlier today.

Thanks for reporting. I fixed the header logo in #400. I think this _just_ happened.

As for not looking good in dark mode, the "stable" build of the docs looks fine to me:
https://python-graphblas.readthedocs.io/en/stable/index.html

The stable build uses `pydata-sphinx-theme` 0.12.0. The "latest" build uses version 0.13.1. Looks like something changed (CC @jim22k). Perhaps we should pin to a specific version.

Also, should we point to "stable" docs instead of "latest" docs? As a developer, it's nice to have issues reported for latest.

Regarding the original issue of a broken link in the doc, perhaps we should add a check to ensure links work. For example, `make linkcheck` mostly does this, but seems to have issues for relative links for me.

SultanOrazbayev:
Just a quick note to confirm the docs render well for me in the dark mode also and I can see the logo.

jim22k:
I just updated readthedocs config so that "stable" is the default.

@eriknw I'll update your PR with a fix for the dark mode.

PR #397, opened 2023-03-07
Update fundamentals.rst

SultanOrazbayev:
Resolve https://github.com/python-graphblas/python-graphblas/issues/396

Checked with `grep -rai '::/'` and seems this is the only case.

### DIFF
--- a/docs/user_guide/fundamentals.rst
+++ b/docs/user_guide/fundamentals.rst
@@ -37,7 +37,7 @@ The descriptor is a set of bitwise flags.
   - Replace mode indicates that elements outside the mask area should be cleared in the final output.
     When not in replace mode, elements outside the mask are left untouched.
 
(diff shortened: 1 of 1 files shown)


eriknw:
Thanks for the quick fix @SultanOrazbayev!

I just fixed the (unrelated) bug, which was a new DeprecationWarning, so can you merge the latest main branch into this PR?

SultanOrazbayev:
Unrelated to the issue/PR, but is it feasible/logical to ignore code tests on the update of docs only?

eriknw:
> Unrelated to the issue/PR, but is it feasible/logical to ignore code tests on the update of docs only?

Yeah, that probably a good idea and doable. Want to give it a try? This might be a good resource:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

PR #400, opened 2023-03-08
Fix path to logo in docs (include `_static` in path)

eriknw:
Our logo just disappeared in the docs. Adding `_static/` to the path seems to fix it. Is `html_static_path` supposed to do this for us?

Reported here: https://github.com/python-graphblas/python-graphblas/issues/396#issuecomment-1458955342

Maybe we should pin to a specific version of `pydata-sphinx-theme`.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -6,6 +6,7 @@
 }
 
 .intro-card {
+    background-color: var(--pst-color-background);
     margin-bottom: 30px;
 }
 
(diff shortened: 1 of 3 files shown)


SultanOrazbayev:
It's odd, but now the dark version is not rendering correctly for me. I tried Safari and Chrome, clearing cache, but it shows now just like in the linked comment (so no logo and white text blocks in dark mode).

PR #413, opened 2023-03-22
Fix right panel visibility overlap issue

### LINKED ISSUE #406: [pyos][docs] In-page TOC overlaps with page content in docs
Not sure if this is an upstream issue with `pydata-sphinx-theme` - but when loading the docs on mobile, or just with a narrow screen, they look like this, which makes them unreadable. Tested in firefox and chrome, though sometimes chrome doesn't show the in-page TOC at all, which is interesting.

If this is an upstream issue with the theme, feel free to close, just wanted to put on your radar

Mobile (simulated):
Screen Shot 2023-03-20 at 11 43 09 AM

Screen Shot 2023-03-20 at 11 43 18 AM

Related to PyOS review: https://github.com/pyOpenSci/software-submission/issues/81

jim22k:
The issue was an entry in the CSS. Not sure where it came from, but it was breaking the behavior of the right-side panel. Once removed, everything works fine.

I also split up the API reference into separate pages for better navigation experience.

Fixes #406


eriknw:
LGTM, thanks @jim22k! I edited your comment to indicate which issue it fixes.

SultanOrazbayev:
Great, the dark theme also looks good now!

Issue #324, opened 2022-11-13
Support Python 3.11


eriknw:
Let's track this. We are waiting on numba:
- https://github.com/numba/numba/pull/8381
- https://github.com/numba/numba/pull/8545
- https://github.com/numba/numba/pull/8590
- https://github.com/numba/numba/issues/8304

`python-suitesparse-graphblas` now supports Python 3.11.

eriknw:
Closing. CI now tests Python 3.11. This was implemented in #423.

PR #414, opened 2023-03-23
Test numba 0.57dev and numpy 1.24

eriknw:
This doesn't yet test with Python 3.11, but it does test the latest CI builds of numba, which lets us also test numpy 1.24.

We expect the numba team to release/announce pre-release versions with Python 3.11 support very soon (perhaps today or tomorrow).

xref #324


eriknw:
This PR is no longer needed; it's now handled in #423

PR #419, opened 2023-03-28
Improve import times

eriknw:
Import times regressed in #388. This PR fixes it.

A nice way to visualize import times is with https://github.com/nschloe/tuna

Now, as before, import times are dominated by numba and numpy.

I looked at timing for "Test Imports" Github Actions to identify what caused the regression:
https://github.com/python-graphblas/python-graphblas/actions/workflows/imports.yml


eriknw:
For example, to use `tuna`:
```
python -X importtime -c "from graphblas import *" 2> import.log
tuna import.log
```

jim22k (approved):

PR #423, opened 2023-04-03
making numba optional

### LINKED ISSUE #415: Make numba able to be optional
The primary reason for this is to support other backends that don't use numba. 

It is also valid to use SuiteSparse:GraphBLAS with limited functionality without numba. Other advantages of not requiring numba:
- able to support newer Python versions faster (see #324)
- able to support PyPy (probably)

I think it's still TBD how we'll support multiple backends and how packages and dependencies will be installed.

Acceptance for closing this issue as complete will be to run tests in CI w/o numba installed and to have good error messages when trying to use features that require numba. Numba may still be listed as a required dependency.

eriknw:
I started poking around to see what it would take to allow numba to be optional. Closes #415. May also address #324.

This is a work in progress and can probably be cleaned up. This also tests against Python 3.11 and PyPy.

Let's see what happens in CI!


graphblas/io.py
```diff
@@ -455,7 +455,7 @@ def to_awkward(A, format=None):
         indices, values = A.to_coo()
         form = RecordForm(
             contents=[
-                NumpyForm(A.dtype.numba_type.name, form_key="node1"),
+                NumpyForm(A.dtype.numba_type.name, form_key="node1"),  # XXX: is numba required?
```
eriknw, line 458: @jim22k, is there a reason `A.dtype.numba_type.name` is used instead of `A.dtype.np_type.name`? Numba doesn't appear to be a strict requirement for Awkward.

eriknw:
PyPy 3.8.16 0_73_pypy had a segfault [here](https://github.com/python-graphblas/python-graphblas/actions/runs/4603607378/jobs/8133767261) when using psg wheel, so I'm skipping PyPy on CI for now (not worth the effort). Most of the PyPy jobs were passing though!

I'm not sure when I'll be able to push this PR over the finish line, but it's reviewable in its current state. Most of the changes are pretty minor, and it was mostly just grunt-work to get tests to pass w/o numba installed.

This PR does not yet introduce "udf" backends or a means to make numba optional when installing.

graphblas/io.py
eriknw: Using `A.dtype.np_type.name` seems to work; I'm going to assume this resolves it and works fine.

eriknw:
Using `numba==0.57.0rc1` appears to work! :tada: (also NumPy 1.24 and Python 3.11)

graphblas/core/operator/binary.py
```diff
@@ -681,21 +713,22 @@ def _initialize(cls):
...
                 new_op._add(cur_op)
-        # Add floordiv
-        # cdiv truncates towards 0, while floordiv truncates towards -inf
-        BinaryOp.register_new("floordiv", _floordiv, lazy=True)  # cast to integer
-        BinaryOp.register_new("rfloordiv", _rfloordiv, lazy=True)  # cast to integer
+        if _supports_udfs:
+            # Add floordiv
+            # cdiv truncates towards 0, while floordiv truncates towards -inf
+            BinaryOp.register_new("floordiv", _floordiv, lazy=True)  # cast to integer
+            BinaryOp.register_new("rfloordiv", _rfloordiv, lazy=True)  # cast to integer
```
eriknw, line 720: For operators that are available by default if UDFs are supported, should we give good error messages such as "numba required..."?

eriknw: Done.

eriknw:
I think this is ready for review/closer scrutiny.

This sets us on the path of having `numba` and even `suitesparse-graphblas` as optional dependencies. Documentation has been updated to show `pip install python-graphblas[default]`.

We may want to wait until numba 0.57 is released to merge this.

jim22k (approved):

eriknw:
I futz around with coverage a _little_ bit, and in the process fixed a few bugs. I also commented code e.g. `# NOT COVERED`, `# BRANCH NOT COVERED`, and `# FLAKY COVERAGE` to call attention.

Anyway, this is going in!  I skip numba 0.57.0rc1 on Windows in CI. We can update CI when there is a new RC, and/or when 0.57.0 is released.

One final reminder: I am curious whether it's better to use `isclose` as a UDF or a recipe. We have both. We should experiment.

PR #441, opened 2023-04-18
Add note about supporting GraphBLAS C API version

### LINKED ISSUE #424: [pyos][question] Handling of graphBLAS API version?
Trying to understand the implementation of the API wrapper, and I'm wondering how the version of the graphBLAS API is handled? It looks like the API moves slowly, but it does move ( eg https://graphblas.org/ says [v1.3.0](https://graphblas.org/docs/GraphBLAS_API_C_v1.3.0.pdf) on September 25, 2019 and [v2.0.0](https://graphblas.org/docs/GraphBLAS_API_C_v2.0.0.pdf) on November 15, 2021 ). 

I ask because it looks like the mappings are relatively manual - a lot of hardcoded names, etc. - rather than something where you're programmatically generating bindings directly from the API. Having just started taking a look at the implementation I am assuming there's a good reason for that, so completely fine, but is there some place where you specify that "this version of python-graphblas is compatible with x version of graphblas API?" 

I see that a version of `suitesparse-graphblas` is pinned, so presumably that also pins some specific version of the API? 

In any case, seems like something that would be worth having in the docs if not implemented in the code - there's a statement on python version here: https://python-graphblas.readthedocs.io/en/stable/getting_started/faq.html#what-is-the-version-support-policy but I don't find one for graphBLAS API version

eriknw:
Closes #424.

This formalizes our policy of how we handle GraphBLAS C API versions. I also added a short note in the README about which version we currently support (I'm not sure how/whether/where to put it in the docs website).

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ or pip:
 $ pip install python-graphblas[default]
 ```
 This will also install the [SuiteSparse:GraphBLAS](https://github.com/DrTimothyAldenDavis/GraphBLAS) compiled C library.
+We currently support the [GraphBLAS C API 2.0 specification](https://graphblas.org/docs/GraphBLAS_API_C_v2.0.0.pdf).
 
 ### Optional Dependencies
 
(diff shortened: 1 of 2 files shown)


README.md
SultanOrazbayev, line 1: Not sure if this adds much value, but could also mention the supported C API version on line 59:

```diff
- The approach taken with this library is to follow the C-API specification as closely as possible while making improvements
+ The approach taken with this library is to follow the C-API 2.0 specification as closely as possible while making improvements
```

eriknw: Done--thanks!

PR #445, opened 2023-04-20
Don't install numba rc on Windows (fix)

eriknw:


eriknw:
nm, it's less flaky w/o this PR. There will be the occasional failure until numba 0.57 is released.

PR #447, opened 2023-04-24
Convert logo from font svg into path svg

jim22k:
Same thing we did for [graphblas-algorithms](https://github.com/python-graphblas/graphblas-algorithms/pull/60) -- change from a small svg which uses built-fonts and instead use an explicit path to remove the font challenges. This has been run thru a minifier.


eriknw:
Yay, thanks @jim22k!

Which minifier did you use? One of the tools I tried when doing https://github.com/python-graphblas/graphblas-algorithms/pull/60 actually lowered the quality of the svg, but I forget which one.

https://vecta.io/nano appears to do a good job and is high quality.

eriknw (approved):

PR #451, opened 2023-05-05
include conftest.py in MANIFEST.in

eriknw:
We somehow forgot to include this. Still need to verify that including this allows us to run tests with custom options, but I think it should.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,7 @@ recursive-include graphblas *.py
 prune docs
 prune scripts
 include setup.py
+include conftest.py
 include README.md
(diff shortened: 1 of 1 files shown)


eriknw:
Okay, so adding `conftest.py` to `MANIFEST.in` actually doesn't add it to the wheel--it's only added to the source distribution. It belongs in the source distribution, so this change is still good.

I just added some logic so that slow tests will run (and not emit warnings) when run via `pytest --pyargs graphblas`.  It also runs "randomly". I don't know how to allow custom cli arguments to be passed in when pytest is invoked in this way.

PR #455, opened 2023-05-17
blacken docs (and other misc)

eriknw:
This doesn't actually add `blacken-docs` as pre-commit hook. `blacken-docs` doesn't like when we sometimes use `>>>` in code blocks, and it doesn't like how we align the code comments in the README.

Also, do a few misc changes pointed out by https://scikit-hep.org/developer/reporeview

### DIFF
--- a/graphblas/viz.py
+++ b/graphblas/viz.py
@@ -35,8 +35,7 @@ def _get_imports(names, within):
             except ImportError:
                 modname = _LAZY_IMPORTS[name].split(".")[0]
                 raise ImportError(f"`{within}` requires {modname} to be installed") from None
-            finally:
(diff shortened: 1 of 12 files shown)


graphblas/viz.py
```diff
@@ -35,8 +35,7 @@ def _get_imports(names, within):
             except ImportError:
                 modname = _LAZY_IMPORTS[name].split(".")[0]
                 raise ImportError(f"`{within}` requires {modname} to be installed") from None
-            finally:
-                globals()[name] = val
+            globals()[name] = val
```
eriknw, line 38: Also, this is a bugfix.

PR #462, opened 2023-06-21
doc: add missing CLI keyword.

SultanOrazbayev:
The previous version did not use the correct syntax for conda environment creation from file.


eriknw:
LGTM, thanks for the attention to detail @SultanOrazbayev!

Btw, I think I fixed the CI failures in #464 (now in main).

Issue #465, opened 2023-06-22
Awkward conversion tests are failing


eriknw:
As seen here (https://github.com/python-graphblas/python-graphblas/actions/runs/5337195569/jobs/9672901538#step:9:731), the tests for `awkward` are failing. I don't know why or when or for what environments they are failing.

<details>

```python-traceback
___________________________ test_awkward_roundtrip ____________________________

array = <Array [{data: [...], ...}, {...}, {...}] type='3 * struct[{data: var * {in...'>
name = '_AwkwardDoublyCompressedMatrix'

    def with_name(array, name, *, highlevel=True, behavior=None):
        """
        Args:
            array: Array-like data (anything #ak.to_layout recognizes).
            name (str or None): Name to give to the records or tuples; this assigns
                the `"__record__"` parameter. If None, any existing name is unset.
            highlevel (bool): If True, return an #ak.Array; otherwise, return
                a low-level #ak.contents.Content subclass.
            behavior (None or dict): Custom #ak.behavior for the output array, if
                high-level.
    
        Returns an #ak.Array or #ak.Record (or low-level equivalent, if
        `highlevel=False`) with a new name. This function does not change the
        array in-place. If the new name is None, then an array without a name is
        returned.
    
        The records or tuples may be nested within multiple levels of nested lists.
        If records are nested within records, only the outermost are affected.
    
        Setting the `"__record__"` parameter makes it possible to add behaviors
        to the data; see #ak.Array and #ak.behavior for a more complete
        description.
        """
        with ak._errors.OperationErrorContext(
            "ak.with_name",
            {"array": array, "name": name, "highlevel": highlevel, "behavior": behavior},
        ):
>           return _impl(array, name, highlevel, behavior)

C:\Miniconda3\envs\graphblas\lib\site-packages\awkward\operations\ak_with_name.py:38: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

array = <Array [{data: [...], ...}, {...}, {...}] type='3 * struct[{data: var * {in...'>
name = '_AwkwardDoublyCompressedMatrix', highlevel = True, behavior = None

    def _impl(array, name, highlevel, behavior):
        behavior = behavior_of(array, behavior=behavior)
        layout = ak.operations.to_layout(array)
    
        def action(layout, **ignore):
[...]

PR #466, opened 2023-06-22
Investigate awkward test failure

eriknw:
Investigate #465


eriknw:
After too much CI testing, the issue is clearly when using awkward with numpy 1.25. I don't understand specifically what the problem is though. But, this is isolated enough for me (for now). Closing.

PR #469, opened 2023-06-23
Clarify in docs that monoids are commutative and associative

### LINKED ISSUE #468: [pyos][docs] "Monoids are also associative, so the order of the inputs does not matter."
[Here](https://python-graphblas.readthedocs.io/en/stable/user_guide/operators.html#monoids) it says,

> Monoids are also associative, so the order of the inputs does not matter.

This is a bit confusing. Did you mean to say that the monoids in py-graphblas are all commutative? Or say that grouping does not matter?

eriknw:
Technically Monoids don't need to commute, but behavior is ill-defined (or invalid) if non-commutative monoids are used for reductions and semirings, so, practically speaking, all monoids should commute in GraphBLAS. I don't know of any use cases for non-commutative monoids--a binary operator should always be possible to use and preferred.

Fixes #468.


SultanOrazbayev (approved):
Looks good, with a possible clarification that GraphBLAS uses commutative monoids (so commutative is not part of the definition of a monoid, but the monoids used by GraphBLAS also have this property). That's my understanding from the "Design of the GraphBLAS API for C" paper and, ahem, Wikipedia.

eriknw:
Thanks for looking @SultanOrazbayev. Refined.

SultanOrazbayev (approved):
Thank you, LGTM!

PR #470, opened 2023-06-24
Update copyright to include contributors

eriknw:
This is like how Dask and Bokeh handle this. @jim22k I'd like your explicit approval.


jim22k (approved):
Looks good to me.

PR #478, opened 2023-06-27
Backport many of the SS:GB 8 changes to run on 7

eriknw:
This takes many changes from #456 and runs against the current set of SuiteSparse:GraphBLAS versions.

I think it would be nice for us to support older versions of SuiteSparse:GraphBLAS if possible (and reasonable), particularly since SS:GB 8.0 seems somewhat experimental.

This also updates to scipy 1.11 and other misc. changes I've been adding to the 8.0 branch.


eriknw:
Dang it. Looks like `fast_matrix_market` and `scipy` are not playing nicely together. Got to figure out the versions, etc.

eriknw:
Okay, I think this is ready to go in.

@jim22k @SultanOrazbayev I'd like to do a release soon that still targets SuiteSparse:GraphBLAS 7 series, so let's try to get our other PRs ready and in. This PR should unblock failures that are popping up due to new dependency versions.

PR #489, opened 2023-08-09
dropping support for python3.8

Transurgeon:
Made changes to dependencies, pyproject.toml and CI workflows according to the instructions listed on the [wiki](https://github.com/python-graphblas/python-graphblas/wiki/Drop-support-of-old-Python-version). 
However, I wasn't sure what to change for this particular instruction:
``- Update minimum dependency versions to match what we test``
``Removing support of a Python version is a natural time to remove supported versions of dependencies, which we may support for much longer than 24 months if it's trivial to do so.``


eriknw:
Thanks @Transurgeon! Welcome as a contributor :tada:

These changes look perfect to me to drop Python 3.8 👍 

For reference, NEP 29 dropped Python 3.8 support in April this year, and SPEC 0 dropped it about a year ago:
- https://numpy.org/neps/nep-0029-deprecation_policy.html
- https://scientific-python.org/specs/spec-0000/

> However, I wasn't sure what to change for this particular instruction:
> > Update minimum dependency versions to match what we test

Aha, it means to update the minimum dependency versions in `pyproject.toml` _if_ removing Python 3.8 from CI means that we no longer test some versions of dependencies. We still test the same versions of dependencies as we did when we tested with Python 3.8, so no changes are necessary this time.

I will update the instructions on the wiki to mention `pyproject.toml`, b/c I agree that wasn't clear. Thanks for the feedback.

> > Removing support of a Python version is a natural time to remove supported versions of dependencies, which we may support for much longer than 24 months if it's trivial to do so.

This means that if we want to, we can also change what versions of dependencies we support. This is completely optional, and I don't want to think about it right now, so let's do nothing :)

Merging!

Issue #487, opened 2023-07-31
Question: Most efficient way to set or remove the diagonal in large matrices


q-aaronzolnailucas:
Hi, what's the best way to set or remove the diagonal (self-loops) from a graphblas Matrix?

eriknw:
Good questions, thanks for asking! Here are a few examples:
```python
>>> import graphblas as gb
>>> import numpy as np

>>> A = gb.Matrix.from_dense(np.arange(9).reshape(3, 3))
>>> A
"M_0"      nvals  nrows  ncols  dtype  format
gb.Matrix      9      3      3  INT64   fullr
---------------------------------------------
   0  1  2
0  0  1  2
1  3  4  5
2  6  7  8

>>> # Diagonal of a Matrix as a Vector
>>> A.diag()
"v_0"      nvals  size  dtype  format
gb.Vector      3     3  INT64    full
-------------------------------------
index  0  1  2
value  0  4  8

>>> # Off-diagonal of a Matrix as a Vector
>>> A.diag(1)
A"v_1"      nvals  size  dtype  format
gb.Vector      2     2  INT64    full
-------------------------------------
index  0  1
value  1  5

>>> # Select diagonal of a Matrix as a Matrix
>>> A.select("diag")
gb.MatrixExpression                         nrows  ncols  dtype
M_0.select(select.diag[BOOL], thunk=False)      3      3  INT64

"Result"   nvals  nrows  ncols  dtype
gb.Matrix      3      3      3  INT64
-------------------------------------
   0  1  2
0  0
1     4
2        8

Do expr.new() or other << expr to calculate the expression.

>>> # Select off-diagonal of a Matrix as a Matrix (using alternative API)
>>> gb.select.diag(A, 1)
gb.MatrixExpression                     nrows  ncols  dtype
M_0.select(select.diag[BOOL], thunk=1)      3      3  INT64

"Result"   nvals  nrows  ncols  dtype
gb.Matrix      2      3      3  INT64
-------------------------------------
  0  1  2
0    1
1       5
2

Do expr.new() or other << expr to calculate the expression.

>>> # Remove diagonal elements of a Matrix using select
>>> A << gb.select.offdiag(A)
>>> A
"M_0"      nvals  nrows  ncols  dtype   format
gb.Matrix      6      3      3  INT64  bitmapr
----------------------------------------------
   0  1  2
0     1  2
1  3     5
2  6  7

>>> # Assign a Vector to the diagonal of a Matrix
[...]

PR #493, opened 2023-08-23
Add `A.setdiag(x, k)`

eriknw:
See: #487.

This adds `setdiag` method to `Matrix` and `Updater` (such as `A("+").setdiag(v)`). `setdiag` is not available on transposed matrices or expressions. `setdiag` is doing an _assignment_, but with functional syntax instead of with `__setitem__`, so it is most similar to setting a Vector or scalar to a single row or column of a Matrix, so we follow the same semantics.

This supports masking and accumulation, but punts for ~complemented masks~ (complemented masks now implemented) or `replace=True`.

~Tests, documentation,~ feedback, and criticism are still needed.


eriknw:
This is ready for review. I updated it to implement option 2 from https://github.com/python-graphblas/python-graphblas/issues/487#issuecomment-1707316715 by adding `mask=` and `accum=` arguments to `setdiag`.

The main remaining question I have is regarding `clear_missing=True` argument. Does it make sense? Is the name good? Or should it not be included at all? Also, are the code and docstrings clear enough, or should any more docs be added?

I think the pros and cons from https://github.com/python-graphblas/python-graphblas/issues/487#issuecomment-1675686422 are still accurate and I think are worth rereading when reviewing this PR. I am +1 for adding `matrix.setdiag` for reasons stated in that comment.

CC @q-aaronzolnailucas @SultanOrazbayev @jim22k

graphblas/tests/test_matrix.py
```diff
@@ -4393,3 +4395,104 @@ def test_power(A):
...
+    # List (or array) is treated as dense
+    A.setdiag([10, 20], 1)
+    expected[0, 1] = 10
+    expected[1, 2] = 20
+    assert A.isequal(expected)
+    # size 0 diagonals (doesn't set anything, but valid call)
+    A.setdiag(-1, 3)
+    assert A.isequal(expected)
+    A.setdiag(-1, -2)
+    assert A.isequal(expected)
```
jim22k, line 4422: I'm confused why these are valid calls. A size 0 diagonal feels like a diagonal out of range.

graphblas/core/matrix.py
```diff
@@ -2805,6 +2805,124 @@ def power(self, n, op=semiring.plus_times):
...
+            If it is Matrix Mask, then only the diagonal is used as the mask.
+        accum : Monoid or BinaryOp, optional
+            Operator to use to combine existing diagonal values and new values.
+        clear_missing : bool, default=True
+            If True, missing elements in Vector values will result in the
+            corresponding diagonal elements to be removed. This parameter has
+            no effect if values is not a Vector or if accum operator is given.
+            The default is True to match normal behavior for assignments, but
+            it requires a call to ``select.offdiag``. If clear_missing is False,
+            then the recipe is approximately ``D = v.diag(k) ; self(D.S) << D``.
```
jim22k, line 2832: I think `clear_missing` should be eliminated. It adds very little utility, adds mental complexity to the API, and the same effect can be achieved with the correctly chosen `accum` (i.e. clear_missing=False is the same as accum=second).

If someone really needed to accumulate diagonal values only where the incoming values are not missing, then they would use a mask. This requires a deeper understanding of how GraphBLAS works, yes, but it achieves the goal of making easy things simple and hard things possible.

graphblas/core/matrix.py
```diff
@@ -2805,6 +2805,124 @@ def power(self, n, op=semiring.plus_times):
...
+        ----------
+        values : Vector or list or np.ndarray or scalar
+            New values to assign to the diagonal. The length of Vector and array
+            values must match the size of the diagonal being assigned to.
+        k : int, default=0
+            Which diagonal or off-diagonal to set. For example, set the elements
+            ``A[i, i+k] = values[i]``. The default, k=0, is the main diagonal.
+        mask : Mask, optional
+            Vector or Matrix Mask to control which diagonal elements to set.
+            If it is Matrix Mask, then only the diagonal is used as the mask.
```
jim22k, line 2823: I like this approach for the mask. We should look into a similar approach for assignment with a row or column if we don't already do this.

jim22k (commented):
Do you plan to allow `A(mask, accum=accum).setdiag(v)`? Which would just delegate to the keywords in the `setdiag` method.

graphblas/tests/test_matrix.py
eriknw: Thanks for paying attention ;)

I flipped a coin for this one, 'cause it could be valid-ish to assign a length 0 vector to a length 0 diagonal, right?

Anyway, I changed the limits to raise for these cases.

graphblas/core/matrix.py
eriknw: Yup, I borrowed this behavior from row/col assign, which already accept matrix masks :)

eriknw:
> Do you plan to allow A(mask, accum=accum).setdiag(v)? Which would just delegate to the keywords in the setdiag method.

Unsure. I removed this. It would be very easy to add back. What do you think?

graphblas/core/matrix.py
eriknw: Oh, right you are! Removed. Much better :)

graphblas/tests/test_matrix.py
jim22k: For a 0-length vector where k == 0, I could maybe see an exception. But if k != 0 and the off-diagonal is length zero, that is out of range. I think raising is the correct approach.

jim22k (approved):

jim22k:
> > Do you plan to allow A(mask, accum=accum).setdiag(v)? Which would just delegate to the keywords in the setdiag method.
> 
> Unsure. I removed this. It would be very easy to add back. What do you think?

Add it as an issue to discuss, but don't hold up this PR for that.

eriknw:
> For a 0-length vector where k == 0, I could maybe see an exception. But if k != 0 and the off-diagonal is length zero, that is out of range. I think raising is the correct approach.

This (allowing to set 0-length when k == 0) is actually what setdiag in scipy.sparse does. I think this is a good idea and let's us better handle length 0 dimensions. For example, should it be possible to set _any_ diagonal in a 0x2 Matrix? My latest commit allows this.

For the record, scipy.sparse setdiag also allows the size of the vector of new values to be too large or too small. Should we follow suite, or require that the vector exactly match the size of the diagonal, which is what we currently do?

graphblas/core/matrix.py
```diff
@@ -2805,6 +2805,119 @@ def power(self, n, op=semiring.plus_times):
...
+
+        if clear_diag:
+            self(mask=mask, **opts) << select.offdiag(self, k)
+
+        Diag = v.diag(k)
+        if Diag.shape != self.shape:
+            Diag.resize(self._nrows, self._ncols)
+        if mask is None:
+            mask = Diag.S
+        self(accum=accum, mask=mask, **opts) << Diag
```
eriknw, line 2919: For posterity and those who browse this PR, these lines are the main recipe for setting the diagonal of a Matrix with a Vector. The rest of the function mostly deals with handing different input types and giving good error messages when necessary.

Issue #492, opened 2023-08-21
Setting or propagating orientation (rowwise or columnwise) of a Matrix


eriknw:
The default orientation of matrices is rowwise. This can be changed globally in SuiteSparse:GraphBLAS with `gb.ss.config["format"] = "by_col"`, and a single matrix can be changed via `A.ss.config["format"] = "by_col"`.

Using matrices with non-default orientation may result in surprising behavior. For example, would users expect elementwise operations to result in the orientation changing (such as `B = (A + 1).new()`)? More illustrations below:
```python
In [1]: import graphblas as gb

In [2]: A = gb.Matrix.from_csc([0, 2, 3, 3, 5], [1, 3, 7, 6, 20], [10, 20, 30, 40, 50])

In [3]: A.ss.format
Out[3]: 'csc'

In [4]: A.ss.orientation
Out[4]: 'columnwise'

In [5]: B = (A + 1).new()

In [6]: B.ss.orientation
Out[6]: 'rowwise'

In [7]: gb.ss.concat([[A, A]]).ss.orientation
Out[7]: 'rowwise'

In [8]: A[:4, :4].new().ss.orientation
Out[8]: 'rowwise'
```
Perhaps this isn't surprising, since a global default means just that: everything is predictable, and _new_ objects by default will use the default orientation. If you want to control the orientation of the output, then this is possible by first creating the output object and setting its orientation.

Some questions:
- should we consider ever propagating the orientation of input matrices to the automatically created result matrix?
- should we consider a way to specify orientation (and other config?) when creating new objects?
  - e.g. `B = (A + 1).new(order="col")`
  - e.g. `B = gb.ss.concat([[A, A]], order="F")`
  - e.g. `B = A[:4, :4].new(order="columnwise")`

It's worth noting that the upcoming GraphBLAS 2.1 specification will introduce the ability to introspect a Matrix for its orientation.

jim22k:
I prefer `.new(order="columnwise")`

Transurgeon:
Hey thanks for bringing this up Erik, I think it would be a great addition.
I also would prefer ``.new(order="columwise")`` but I think all three options are equally suitable for this.

How difficult would it be to propagate the orientation of the input matrices to the result matrix? I feel like this is how it should work (intuitively) anyways. (as in, if someone cared enough about orientation to set a matrix to col, it should probably propagate)

Could you provide some pointers on how one could get started on adding this? I would love to give it a go.

[...]

PR #495, opened 2023-08-26
Adding matrix ``order=None`` to matrix classes

Transurgeon:
This PR adds an optional order argument to some matrix classes.
This is also the first task of issue #492

### DIFF
--- a/graphblas/core/expr.py
+++ b/graphblas/core/expr.py
@@ -486,7 +486,7 @@ class InfixExprBase:
         self.right = right
         self._expr = None
 
-    def new(self, dtype=None, *, mask=None, order=None, name=None, **opts):
+    def new(self, dtype=None, *, mask=None, name=None, **opts):
         if (
             mask is None
             and self._expr is not None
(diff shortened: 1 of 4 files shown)


eriknw:
Thanks for starting on this @Transurgeon!

I would recommend adding implementation too. Ideally, you should test and verify what you did is needed and works as expected as you add `order=` arguments. Perhaps this also means adding `my_matrix.orientation` attribute too, or just using the existing `my_matrix.ss.orientation` attribute.

I would start with `A.T.new(order="columnwise")` and test that it works (add tests at the end of `graphblas/tests/test_matrix.py`).

Also, `graphblas/core/utils.py:get_order` will be helpful.

Transurgeon:
Hello @eriknw , sorry for the delay on this, I have been finishing up some details for my GSoC project. 
I tried to implement the orientation today, however, it doesn't seem to be working properly. 
I have added the argument to both ``Matrix.__new__`` and ``TransposeMatrix.new()``. I also called ``get_order`` like you recommended, but the test that I have added doesn't seem to pass like I would expect. 
I am starting to get lost in the layers of OOP abstraction. 
More importantly, I am also having problems when debugging due to the error below, do you have any ideas on how to fix this?
image
P.S Is the expectation to add ``property.getter/setter`` for ``.ss.orientation``? I thought that could be a solution but wasn't sure if it was what you wanted.

Issue #498, opened 2023-08-30
Correctness Bug: wrong semiring used in functional-style triple product


alugowski:
The functional style multiply, i.e. `gb.semiring.plus_plus(A @ B) `, appears to not respect the semiring choice when doing a triple product `A @ B @ C`.

Let:
```
A = gb.Matrix.from_coo([0, 0, 1], [0, 1, 1], [11, 22, 33], nrows=2, ncols=2)

left = gb.Matrix.from_coo([0, 0], [0, 1], [1, 1], nrows=1, ncols=2)
right = gb.Matrix.from_coo([0, 1], [0, 0], [1, 1], nrows=2, ncols=1)
```

A triple product using the `plus_first` semiring returns a 1x1 matrix with a count of nonzeros in A, which is 3. Indeed that is what the method style triple product returns:
```
left.mxm(A, op="plus_first").mxm(right, op="plus_first")
# returns `[[3]]` (correct)
```

However the functional-style triple product,
```
gb.semiring.plus_first(left @ A @ right)
# returns `[[66]]` (incorrect)
```

It appears that the `plus_first` semiring is only used for one of the multiplications, and `plus_times` for the other.

Application is to implement a triple product that creates a low resolution spy plot, like this:

<img width="400" alt="triple_product" src="https://github.com/python-graphblas/python-graphblas/assets/2730364/fb7ed128-0f1c-4bd6-8c0c-972570d1b929">

See [MatSpy](https://github.com/alugowski/matspy).

eriknw:
Hey @alugowski, thanks for raising this usability issue! I hope you raise more suggestions :)

I'm sorry for my delay--I was traveling and largely AFK last week.

This issue has come up before, and you can see our previous discussion in https://github.com/python-graphblas/python-graphblas/issues/132.

I'm open to adding this functionality, but I want us to think hard about it. Please be patient as we seek consensus (including yours!) and explore options and implementations.

I think the behavior you expect for matrix multiply is very reasonable, and is a pattern I expect to occur semi-regularly (you're not the first to encounter this). Our current behavior may be surprising and counterintuitive--especially for users coming from `pygraphblas`.

I have begun an implemention in #501.

alugowski:
No worries!

[...]

Issue #132, opened 2021-12-01
`AA << op.plus_times(P.T @ A @ P)`  ?


ParticularMiner:
Hi @eriknw 

Oops!  I stretched the infix notation a bit too far by testing the following line, which is an expression one would often encounter when transforming matrix-quantities (like graph adjacency-matrices, for example):

```python
AA << op.plus_times(P.T @ A @ P)
```

where `A`, `AA` and `P` are all matrices.  (That is, `A` is transformed by `P` into `AA`.  `P` could be a permutation matrix, for example.)

The expression was executed without throwing an exception, but the result was incorrect.  To be precise, the structure of the answer was correct but the values were all 0 (which was incorrect).

Of course, the right way to do it was this:

```python
AA << op.plus_times(P.T @ op.plus_times(A @ P))
```
which gave the right answer.

But it would be great if the former notation was also possible.  What do you think?

ParticularMiner:
@eriknw 

Sorry, I made a mistake.  Both expressions work as expected.   I was using the wrong operator.  It should have been `op.plus_times` not `op.plus_minus`.   Obviously!

ParticularMiner:
Surprisingly, it works!

eriknw:
Good question.  Yes, I expect this to work *only* for the `plus_times` semiring.

- `AA << A @ P` is equivalent to `AA << op.plus_times(A @ P)`.
- `C << A | B` is equivalent to `C << op.lor(A | B)` *only* if `A` and `B` are BOOL (otherwise, this raises).
- `C << A & B` is equivalent to `C << op.land(A & B)` *only* if `A` and `B` are BOOL (otherwise, this raises).

Although I generally prefer to be explicit, we think these are convenient enough to be special.  The above behavior can be disabled via `grblas.config.set(autocompute=False)` (which may also require you to call `.new()` more often).

Hence, `AA << P.T @ A @ P` is equivalent to `AA << op.plus_times(op.plus_times(P.T @ A).new() @ P)`.

Note that this doesn't extend to other semirings.  For example, `AA << op.min_plus(P.T @ A @ P)` is equivalent to `AA << op.min_plus(op.plus_times(P.T @ A).new() @ P)`.

ParticularMiner:
Makes sense.  

So `op.plus_times` is the default semiring for the `@` operator.
And similarly for the other semirings and operators you mentioned.

Then your last line:

[...]

PR #501, opened 2023-09-05
Add `semiring(A @ B @ C)` that applies semiring to both matmuls

eriknw:
This is an experiment and possible beginning of implementation to address #132 and #498. This allows a semiring to be applied to many matrix multiplies, such as `gb.op.plus_plus(A @ B @ C @ D)`.

I believe this pattern does arise in practice, and I expect this to be used in a way that is readable and intuitive. Currently, this:
```python
D << gb.op.plus_plus(A @ B @ C)
```
is equivalent to:
```python
D << gb.op.plus_plus(op.plus_times(A @ B) @ C)
```
which may be surprising and counterintuitive.

TODO:
- [x] tests
- [x] document (where/how?)
- [x] seek consensus
- [x] decide if we want to do similar for ewise infix such as `binaryop(A | B & C)` (**do not support for now**)


eriknw:
Hey @jim22k @SultanOrazbayev @michelp, this is pretty much ready if you want to take a look--I just need to write the error message when mixing infix notations such as `binop(x & y | z)`.

This allows operators for ewise add, ewise mult, and matrix multiplication to apply to all the operands inside the operator when using infix notation such as `plus(a | b | c | d)` or `plus_plus(A @ B @ C)`. This doesn't do any fancy reordering, applying of masks, or reuse of intermediates--it simply automatically computes the intermediate values (regardless of `gb.config["autocompute"]` setting).

Mixing ewise infix notation now raises: `binop(x & y | z)` is not allowed. This is because I think many people get tripped up by the operator precedence of `&` and `|`. The only time this may be a nuisance is if you are dealing with boolean values and are comfortable using `&` and `|` together, but I think it's better for the sake of future readers to require explicit `lor` and `land` as appropriate.

Applying operators like this also _only_ apply to infix notation. `plus(a | b | c)` is not the same as `a.ewise_add(b | c, plus)` or `(a | b).ewise_add(c, plus)` (both of which are kind of weird and will probably raise).

Overall I think this change is for the better. I think the main "gotcha" this introduces is this:
```python
>>> C = A @ B
>>> E << min_plus(C @ D)
```
Did they mean `C = min_plus(A @ B)` or `C = plus_times(A @ B)`? This could be avoided many different ways (depending on what is intended):
1. `C = (A @ B).new()`
2. `C << A @ B`
3. `C = semiring(A @ B)`
4. `C << semiring(A @ B)`
5. `expr = A @ B  # Using a better name to indicate an operator may still be applied`

Hence, it's best practice to either be explicit with a semiring, call `.new()`, or assign into an output via `<<`.

Unless we want to detect whether an expression is assigned or inlined (which is possible, but let's not introduce that voodoo magic if we don't absolutely need to), this is the tradeoff we are stuck with. I think the multi-infix notation can be used well, as @alugowski did in #498. Edge cases exist here no matter what we do (I think), and I'm inclined to support what multiple users expected to work, which is what this PR does. This PR goes further by detecting and disallowing other syntax that is likely to be incorrect.

When we discussed this possibility before in #132, @ParticularMiner liked applying the operator to all infix operands as done in this PR (see: https://github.com/python-graphblas/python-graphblas/issues/132#issuecomment-984129848)

eriknw:
This PR should now be easier to review b/c I added unrelated changes into main via #517.

jim22k:
I'm okay with this approach. I agree that `gb.min_plus(A @ B @ C)` is more likely than `X = A @ B; gb.min_plus(X @ C)` and there will be some confusion either way. Such is a nature of delayed computation systems.

eriknw:
This is in--thanks all!

I found it nice to use this syntax when creating #518 (I wrote `semiring(I @ A @ I)`)

PR #503, opened 2023-09-07
fixing broken link to graphblas.org

Transurgeon:
updated a link to graphblas.org in the primer to point to the correct address.


eriknw:
Good catch--thanks @Transurgeon!

PR #508, opened 2023-10-08
Add support for Python 3.12

eriknw:
Also, test against python-suitesparse-graphblas 8.2.0.1

I followed the instructions: https://github.com/python-graphblas/python-graphblas/wiki/Support-new-Python-version


eriknw:
I think this is very close, but for sanity, I'd like to wait until matplotlib has been released and built for Python 3.12.

eriknw:
Thanks @galaxy001, right you are--good timing! Plus, Python 3.12 support just arrived this morning for matplotlib on conda-forge.

Now we just need to wait for tests to pass :)

Another things this PR portends is that we will likely stop using coveralls.io soon and rely only on codecov.io for coverage services. coveralls does not appear to be well-maintained for Python, and it restricts our version of coverage, which may be causing other headaches.

PR #510, opened 2023-10-14
Add NumFOCUS badge

eriknw:
Where should we say Python-graphblas is a NumFOCUS affiliated project? Also, could/should we add an Affiliations section where we also mention Anaconda, Intel, DARPA, NVIDIA, or whatever? This probably requires asking permission, so meh.

I've seen other affiliated projects use the "powered by NumFOCUS" badge like this.


eriknw:
This moved some badges around. Dunno if anybody has strong feelings about what goes where.

SultanOrazbayev:
HoloViz also has a badge (like you added here), so that looks good.

I guess there will eventually also be a link from here: https://numfocus.org/sponsored-projects?_sft_project_category=python-interface (right now there is no link to python-graphblas)

eriknw:
Yeah, there will eventually be a link from here:
https://numfocus.org/sponsored-projects/affiliated-projects
We're not making this up; they're just busy!

PR #512, opened 2023-10-15
Add docstrings for using SS JIT, and make better

eriknw:
Also:
- allow a SS JIT function to be registered under the same name with different input types
- be extra-strict about input types for SS JIT
- test numba JIT with select with udt

CC @michelp in case you want to look at the docstrings and signatures of our functions for using SuiteSparse:GraphBLAS JIT such as `gb.binary.ss.register_new`--I think you're the most likely to show off something cool using the JIT ;)
(also, let me know if you want help using the JIT locally).

I think we should cut a new release after this is merged.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,6 +8,7 @@ coverage:
         informational: true
     changes: false
 comment:
+  layout: "header, diff, footer"
   behavior: default
 github_checks:
   annotations: false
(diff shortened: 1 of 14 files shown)


michelp (approved):
lgtm!

eriknw:
Thanks for taking a look @michelp!

So I added codecov github comment in this PR. I don't know if I like it, but it's convenient to have an easy link to the coverage report. I made the message not quite as ugly as the default.

Anyway, merging! I'll get overall approval from others before I do a release, which is coming soon.

PR #515, opened 2023-10-19
Try using `sysconfig` to configure the SS JIT

eriknw:


eriknw:
Lessons learned:
- Using `sysconfig` to provide defaults appears to "just work" for Ubuntu and Mac
  - It also appears to mostly work on Windows except when using `python-suitesparse-graphblas` wheels (which may not play nicely with conda-forge)
- pandas 2.1 and numpy 1.21 can't be installed together (note to myself)

So, using `sysconfig` seem like sane defaults to me (much more sane than using build-time info). How should we set this for Python users? Should we set this by default in `python-suitesparse-graphblas`, and, if so, should we save the original config and make it easy to restore?

PR #519, opened 2023-11-05
Add `gb.MAX_SIZE`, which is `GrB_INDEX_MAX + 1`

eriknw:
`INDEX_MAX` may be a nice global value to expose. The GraphBLAS spec doesn't have many of these globals.

This is the largest index value, not the size of the domain of indices. For example, create the largest vector via `Vector(int, INDEX_MAX + 1)`. Also, note that the spec does not specify a value for `GrB_INDEX_MAX`, only that implementations define it, so it may be backend-dependent.

Alternatively, we could instead define `MAX_SIZE`, which would be `GrB_INDEX_MAX + 1`. I think this is much more natural to use; e.g., `Matrix(int, MAX_SIZE, MAX_SIZE)`. I don't think we need both `INDEX_MAX` and `MAX_SIZE`.

So, which to y'all prefer, `INDEX_MAX` or `MAX_SIZE`? Feel free to suggest different names too. 
CC @michelp @jim22k @SultanOrazbayev

### DIFF
--- a/graphblas/__init__.py
+++ b/graphblas/__init__.py
@@ -39,6 +39,7 @@ del get_config
 backend = None
 _init_params = None
 _SPECIAL_ATTRS = {
+    "INDEX_MAX",
     "Matrix",
     "Recorder",
     "Scalar",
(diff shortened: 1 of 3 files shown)


michelp:
+1 for MAX_SIZE

michelp (approved):

SultanOrazbayev (approved):
LGTM!

eriknw:
Thanks, this is in!

PR #537, opened 2024-02-07
Adopt SPEC 0

eriknw:
More projects are endorsing/adopting SPEC 0, and I think we should too:
- https://scientific-python.org/specs/spec-0000/

Even NumPy is moving towards SPEC 0 instead of [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

Basically, we should feel okay moving forward and dropping old dependencies. This reduces our maintenance burden, and reduces maintenance burden for the whole ecosystem as SPEC 0 gets more widely accepted.

Changes:
- Drop Python 3.9
- Drop NumPy 1.22
- Drop a few other minor dependencies

@jim22k @SultanOrazbayev, since this is a policy change, I want your approval (even a thumbs up or rocket would suffice).


SultanOrazbayev:
Sounds very reasonable.

PR #546, opened 2024-07-01
Update to support latest versions, including NumPy 2

eriknw:
Notably, `python-suitesparse-graphblas` still needs updated to support NumPy 2


SultanOrazbayev:
Perhaps worth waiting for 2.1 or the earliest version that takes into account any compatibility/transition issues?

eriknw:
Okay, I think this is ready 🚀 !

PR #557, opened 2025-02-26
Add 9.4.5.0 to tests

jim22k:


eriknw (approved):
Thanks @jim22k, looks great! I ran CI twice for sanity.

#546 made these branches of setting `psgver` a bit more clear by doing each Python version and source or upstream, but it does add some duplication and additional effort to add new versions. Hopefully it's not too bad. Simple is probably better than fancy here if we look at it every 6-12 months.

eriknw:
Also worth noting that we don't need a new release to update metadata since the latest release uses this constraint:
https://github.com/python-graphblas/python-graphblas/blob/22d42f615187a3a2c89b6400a898c5b72ef396a8/pyproject.toml#L75

PR #570, opened 2026-03-21
Fix segfault when UDT name >= GxB_MAX_NAME_LEN

jim22k:
Claude to the rescue after not understanding the root cause.


eriknw:
Looks great, thanks @jim22k (and Claude)! The failing CI is expected--we haven't updated things in a while. Merging

PR #588, opened 2026-08-04
Compile UDFs under Numba's numpy error model

eriknw:
A UDF that divides by zero silently corrupted the result. Numba's default
error model raises ZeroDivisionError inside the cfunc, where Numba prints
the traceback and returns without writing, so GraphBLAS keeps whatever the
output element already held. Measured at 6f1eb02:

    def _idiv(x, y):
        return x // y
    op = BinaryOp.register_anonymous(_idiv, "_probe_idiv")
    op(Vector.from_coo([0, 1], [10, 20]) & Vector.from_coo([0, 1], [2, 0])).new()
    -> [5, 5]        # numpy gives [5, 0]; the 5 is the previous element

The float case is the same shape: `[0.5, 0.5]` where numpy gives
`[0.5, inf]`. The only user-visible signal is an "Exception ignored"
traceback on stderr.

Pass `error_model="numpy"` at every `numba.njit` and `numba.cfunc` site for
user-defined ops. Setting it on the cfunc alone is not enough: `_build`
calls `.compile(sig)` on the Dispatcher first, a Dispatcher keeps one
compilation per signature, and whichever compile happens first fixes the
model. Reverting only the Dispatcher argument, leaving it on the cfunc,
reproduces `[5, 5]`.

This changes what a dividing UDF returns, but the behaviour it replaces is
an unwritten element rather than an error, so there is no correct result
being taken away.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


eriknw:
This was a good catch by my agents! The "numpy" error model is indeed what we want, not the "python" error model that tries and fails to raise an exception. Merging before the latest CI can complete (it was green before, and this let's us review 589 sooner, which fixes the segfault in CI)

PR #626, opened 2026-08-06
Pick CI's random matrix values with bash instead of a marketplace action

eriknw:
The rngs gate jobs and every build_and_test job resolved
ddradar/choose-random-action before starting, and GitHub's action
resolution service has been failing that lookup with repeated 5xx
errors ("Internal Server Error occurred while resolving
ddradar/choose-random-action@v4.1.0", also plain Service Unavailable).
A failure in a rngs job cancels everything gated behind it: the whole
11-job matrix in test_and_build.yml and the test_imports job in
imports.yml, which triggers on every pull request. Ten of the last ten
red Test runs died exactly there before running anything.

All six RNG steps (four in test_and_build.yml, two in imports.yml)
drew uniformly from small lists, which bash can do by itself: index an
array with RANDOM modulo its length. RANDOM is a bash builtin, so it
works in the ubuntu default shell, git-bash on Windows, and the bash
3.2 that macOS runners provide, and the jobs no longer download
anything to make their picks. Weighting a choice is now expressed by
repeating it in the list, and skipping one by removing it, which
replaces the action's natural-number weights mechanism (the old
comment about setting weights very large to skip an entry is gone with
it).

Verified: both workflows parse as YAML, the snippet produces uniform
draws under bash 3.2.57 (macOS's /bin/bash, with -e -u -o pipefail),
and no uses: of the action remain in any workflow.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/.github/workflows/imports.yml
+++ b/.github/workflows/imports.yml
@@ -13,32 +13,22 @@ jobs:
       os: ${{ steps.os.outputs.selected }}
       pyver: ${{ steps.pyver.outputs.selected }}
     steps:
+      # Plain bash instead of a marketplace action; see the note on the rngs job in
+      # test_and_build.yml. To weight a choice, repeat it; to skip one, remove it.
       - name: RNG for os
-        uses: ddradar/choose-random-action@v4.1.0
         id: os
-        with:
-          contents: |
-            ubuntu-latest
-            macos-latest
-            windows-latest
-          weights: |
-            1
-            1
-            1
+        run: |
+          choices=(ubuntu-latest macos-latest windows-latest)
+          selected=${choices[$((RANDOM % ${#choices[@]}))]}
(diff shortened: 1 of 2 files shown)


eriknw:
LGTM, but should probably wait until CI runs for this before merging. The queue has been pretty backed up, but that's okay, we're patient. Our randomized CI has been earning its keep: we've been finding and fixing things, but this has led to the stacked PRs having a lot of CI jobs. Let's call this an exercise in confidence!

As with some other agent-assisted PRs, the comments can be a little verbose, but I may do a pass at improving them for all PRs in this stack in... a new PR for this stack.

PR #629, opened 2026-08-26
Cover SuiteSparse:GraphBLAS 10.4 and 10.5 in the CI psg pool

eriknw:
Stacked on #628. Second rung below the pre-existing stack; sits between #628 and #587.

The pool stopped at psg 10.3.1.0, so CI installed nothing newer except on the draw that picks "latest", which does not pin and so never produces a reproducible combination. Three releases have landed since, and each carries a SuiteSparse:GraphBLAS this project has to work against: 10.4.0 moved a field in the matrix struct, and 10.5.0 dropped the post-build iso check that four tests used to rely on.

psg 10.5.0.0 is on PyPI and, since 2026-08-26, on conda-forge, so the conda-forge, wheel and source draws all cover it.

`scripts/check_versions.sh` tracks the newest version we test, so its psg floor moves with the pool.

Only the psg pools change here. The other pools are revised further up the stack (#631), so raising them belongs there rather than in a commit that would conflict with every rebase.

Note on placement: this sits BELOW rung 07 so that every rung's CI can draw SuiteSparse 10.4/10.5 against the adaptations in #628. Until #628 and this PR merge into main, #587's Files tab also shows these two commits (the stack manager would not allow retargeting #587's base); it self-heals once these merge.


eriknw:
The CI failure is a segfault that should be fixed by #589 (once we get to that one). This PR is foundational and improves how dependencies are handled in CI. Merging.

